PR #4, opened 2024-09-01
[DRAFT] Adding Ports - WIP

tanneberger:


include/reactor-uc/port.h
```diff
@@ -5,23 +5,43 @@
 #include "reactor-uc/reaction.h"
 #include "reactor-uc/reactor.h"
 #include "reactor-uc/trigger.h"
+#include "reactor-uc/environment.h"
 
 typedef struct InputPort InputPort;
 typedef struct OutputPort OutputPort;
 
 struct InputPort {
-  Trigger super;
```
erlingrj, line 13: The general pattern I went for is to have the super-class be the very first field of the struct such that it can be casted to the super class directly. Is there a good reason for not adhering to this pattern in the case of ports?

include/reactor-uc/reactor.h
```diff
@@ -11,17 +12,21 @@ typedef struct Reaction Reaction;
...
+  // member variables
+  Environment *env;
   Reactor **children;
   size_t children_size;
   Reaction **reactions;
   size_t reactions_size;
   Trigger **triggers;
   size_t triggers_size;
-};
+} __attribute__ ((aligned (32)));
```
erlingrj, line 27: Why do we need to specify the alignment here?

include/reactor-uc/scheduler.h
```diff
@@ -1,18 +1,25 @@
...
-  Environment *env;
-  EventQueue event_queue;
-  ReactionQueue reaction_queue;
+  // methods
   void (*run)(Scheduler *self);
-  void (*prepare_timestep)(Scheduler *self);
+  void (*prepare_time_step)(Scheduler *self);
+
+  // member variables
+  Environment *env_;
```
erlingrj, line 18: The underscore postfix is to signal that these are private fields which should not be modified directly? I guess this is a fair convention. I think though most (all?) of the fields of these structs would qualify as private since they are only changed in the member methods, if this is the case then I propose to not add this postfix

src/builtin_triggers.c
```diff
@@ -1,4 +1,4 @@
-#include "reactor-uc/builtin_triggers.h"
```
erlingrj, line 1: Is this necessary?

src/hardware.c
erlingrj, line 1: This is the style currently pursued by reactor-c. I don't love the conditional inclusion of source files. I was thinking to do this more object oriented and define a `struct Hardware` which defines a set of methods (e.g. wait_until etc) and then do the inheritance trick for each platform. E.g:
```
struct RiotHardware {
  Hardware super;
  ...
}

```

Then we instantiate a `HardwareRiot` and  add a pointer to it in the Environment?

What do you think?

src/hardware/posix/posix_implementation.c
```diff
@@ -0,0 +1,22 @@
+#include "reactor-uc/hardware_interface.h"
+
+#include <time.h>
+
+void hardware_wait_for(instant_t current_time, instant_t wake_up_time) {
```
erlingrj, line 5: WOuld rename to `hardware_wait_until`

erlingrj (changes requested):
This is a nice starting point, there are a few questions to resolve. Also we need to figure out where the values communicated over ports are to be stored and how e.g. the `set` function on an output port moves the value from the output-port and through the connection and to all downstream ports. 

Also we must keep the possibility of delayed and enclaved connections in mind. I don't remember now how I was planning to do it hehe

PR #5, opened 2024-09-17
Connections, Ports and remove generics

erlingrj:
- Add simple Connection primitive. Currently only support for OutputPort -> InputPort, we must also think about Output->Output and Input->Input. Maybe also banks/multiports
- Remove the generics stuff. It was a bad idea
- Add InputPort and OutputPort classes and an example program using them
- Add LogicalAction class and port the example over to it.
- Add `lf_set` and `lf_schedule` to macros.


erlingrj:
@tanneberger I would love to merge your changes with this, especially the level calculations and RIOT support.

PR #16, opened 2024-09-27
Use Justfile instead of Makefile for shorthand commands

erlingrj:
So we can use the Makefile for integration into RIOT and other Makefile based frameworks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,14 +20,14 @@ jobs:
         with:
             submodules: recursive
       - name: Install deps
-        run: sudo apt-get install lcov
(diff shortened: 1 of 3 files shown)


erlingrj:
> @erlingrj whats the benefit over using just classic make?

Because I want to reserve the Makefile for RIOT integration. I.e. `Makefile` should include build info like specifying all sources, include paths etc, such that we can easily include this into an existing project.

We could of course also put that in `reactor-uc.mk` and keep the existing Makefile as is. Maybe a bit confusing

tanneberger:
how about just naming it `reactor-uc-riot.mk` then its clear what its used for.

PR #30, opened 2024-10-02
Connections as trigger and a comprehensive restructuring

erlingrj:
- Redesign of the Trigger type hierarchy
- Connections are now Triggers. This is a good abstraction for Delayed, Physical and Federated connections.
-


erlingrj (commented):
Self-review looks good.

PR #32, opened 2024-10-03
Memory report ci

magnmaeh:
This PR adds a workflow that comments a PR with the change in memory usage for all the C tests. An example of how it will look can be seen here: https://github.com/erlingrj/reactor-uc/pull/56. From the results of the memory report, it is clear the .bss memory region has been massively increased.

It compares the memory usage of the target branch with the head branch to calculate an increase percentage. In doing so it checks out the target branch and runs an action there. Because of this, the CI fails for this PR because the main branch does have the memory-report action yet. But because it works on https://github.com/erlingrj/reactor-uc/pull/56 I expect it will work if we merge this to main as well.


magnmaeh:
> Thanks for addressing this. Will be a very useful feature! However, now the artefacts are committed directly to the main branch of the repo.

Oh okay, I will see if I can figure out how to fix that. I'm very new to this haha

magnmaeh:
I still need to figure out your comment Erling but now it's generating a human-readable report. I guess an improvement would be to format it as a table and add it directly as a comment to the PR. That would be pretty cool actually.

erlingrj:
> I still need to figure out your comment Erling but now it's generating a human-readable report. I guess an improvement would be to format it as a table and add it directly as a comment to the PR. That would be pretty cool actually.

Yes! Would be great to just have it in the comments yes, and a table like for the coverage would be amazing. Btw, the numbers report quite big changes e.g. 50% reduction in text size for many test, however this PR doesnt not introduce any code changes?

magnmaeh:
Ah yes at this point I'm just testing with some random values I set in some files. Just to see what it might look like. Will remove it later

magnmaeh:
I think it's ready for review now! See the description on the PR for why the CI job is failing.

erlingrj (approved):
This looks great! I think this is going to be very useful! In the future I want to do the report based on compilation targeting our embedded platforms.

PR #40, opened 2024-10-05
Merge in main and some restructuring

erlingrj:
1. Merge with main
2. Add nanopb as a gitmodule and place it under `external`
3. Use nanopb's cmakelists 
4. Put proto definition and generated sources/headers under `external/proto`. We don't want to run clang-format on them
5. Address some compiler warnings


tanneberger:
cherry-picked into #41

PR #49, opened 2024-10-09
Fixing couple of loose ends TCP/IP

tanneberger:


src/platform/posix/tcp_ip_bundle.c
```diff
@@ -83,22 +84,25 @@ lf_ret_t TcpIpBundle_send(TcpIpBundle *self, PortMessage *message) {
...
   // serializing protobuf into buffer
-  int status = pb_encode(&stream, PortMessage_fields, message);
+  int message_size = encode_protobuf(message, self->write_buffer, TCP_BUNDLE_BUFFERSIZE);
 
-  if (status < 0) {
+  if (message_size < 0) {
     return LF_ERR;
   }
 
   // sending serialized data to client
```
erlingrj, line 94: It can also be sending the data from the client to the server?

include/reactor-uc/platform/posix/tcp_ip_bundle.h
```diff
@@ -7,6 +7,7 @@
 #include "reactor-uc/error.h"
 
 #define TCP_BUNDLE_BUFFERSIZE 1024
+#define TCP_IP_TIMEOUT 255;
```
erlingrj, line 10: I would propose to call it `TCP_IP_NUM_RETRIES`

src/platform/posix/tcp_ip_bundle.c
```diff
@@ -83,22 +84,25 @@ lf_ret_t TcpIpBundle_send(TcpIpBundle *self, PortMessage *message) {
...
-  ssize_t bytes_written = write(socket, self->write_buffer, stream.bytes_written);
+  ssize_t bytes_written = 0;
+  int timeout = TCP_IP_TIMEOUT;
+
+  while (bytes_written < message_size && timeout > 0) {
+    bytes_written += write(socket, self->write_buffer + bytes_written, message_size - bytes_written);
+    timeout--;
+  }
```
erlingrj, line 101: This also seems to bound the size of the messages we can write? Should the timeout really be decremented unless `write` returns `-1`?

src/platform/posix/tcp_ip_bundle.c
```diff
@@ -83,22 +84,25 @@ lf_ret_t TcpIpBundle_send(TcpIpBundle *self, PortMessage *message) {
...
 
-  // checking if the whole message was transmitted
-  if ((size_t)bytes_written < stream.bytes_written) {
-    return LF_INCOMPLETE;
+  // checking if the whole message was transmitted or timeout was received
+  if (timeout == 0) {
```
erlingrj, line 104: We might have succeeded at the last try, so should check `bytes_written` also

src/platform/posix/tcp_ip_bundle.c
```diff
@@ -141,13 +145,16 @@ PortMessage *TcpIpBundle_receive(TcpIpBundle *self) {
...
 
   self->read_index += bytes_read;
-  pb_istream_t stream = pb_istream_from_buffer(self->read_buffer, self->read_index);
 
-  if (!pb_decode(&stream, PortMessage_fields, &self->output)) {
-    printf("decoding failed: %s\n", stream.errmsg);
+  int bytes_left = decode_protobuf(&self->output, self->read_buffer, self->read_index);
+
+  if (bytes_left < 0) {
     return NULL;
```
erlingrj, line 152: I think we want to change the function type to return `lf_ret_t` and take a `PortMessage**` as an argument so we can get some info back, e.g. that the other side disconnected

tanneberger: I dont think there is anything to gain from returning `lf_ret_t` because, there are only to possible responses `LF_OK` or `LF_ERR` I don't get more detailed feedback. So I would just stick with this more simplistic interface.

PR #59, opened 2024-10-14
Remove unistd and ioctl from TcpIpChannel

LasseRosenow:
This changes the TcpIpChannel implementation in the following way
- `ioctl.h` => `sys/select.h`
- `unistd.h` => `sys/socket.h`

This PR is a draft because I didn't have time to test it yet. There might be implementation bugs.

(I had some problems with compiling the example applications in examples/posix)

### DIFF
--- a/src/platform/posix/tcp_ip_channel.c
+++ b/src/platform/posix/tcp_ip_channel.c
@@ -8,10 +8,8 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <sys/ioctl.h>
 #include <sys/socket.h>
-#include <unistd.h>
-
+#include <sys/select.h>
 #include <nanopb/pb_decode.h>
 #include <nanopb/pb_encode.h>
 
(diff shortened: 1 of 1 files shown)


tanneberger (approved):

src/platform/posix/tcp_ip_channel.c
```diff
@@ -142,25 +140,25 @@ TaggedMessage *TcpIpChannel_receive(NetworkChannel *untyped_self) {
...
+  fd_set read_fds;
+  struct timeval timeout;
+  FD_ZERO(&read_fds);
+  FD_SET(socket, &read_fds);
+  timeout.tv_sec = 0;
+  timeout.tv_usec = 0;
+
+  // check if data is available
+  int select_result = select(socket + 1, &read_fds, NULL, NULL, &timeout);
+  if (select_result <= 0) {
```
erlingrj, line 153: Do we actually need to check how much data is there before we read? Since this is running in a separate thread, we want it to block on `recv` and return with as much data is available. What is gained from checking how much data is there before the read? 

(I know that this was there before your edits, and perhaps @tanneberger can answer better?)

erlingrj:
Great to see that you are already contributing code Lasse. I think the CI ight fail because the PR is coming from your fork. Could you try just filing it from a branch in this repo. I think you have the necessary access.

LasseRosenow:
It seems you are correct, I will close this in favor of #61

PR #61, opened 2024-10-15
Remove ioctl from TcpIpChannel

LasseRosenow:
This changes the TcpIpChannel implementation in the following way
- `ioctl.h` => `sys/select.h`
- Use unistd.h for `close` instead of `shutdown`, which is not available in `RIOT`

The rationale behind this is to improve compatibility with platforms such as `zephyr` or `RIOT` which don't have full POSIX support.

This PR is a draft because I didn't have time to test it yet. There might be implementation bugs.

(I had some problems with compiling the example applications in examples/posix)


src/platform/posix/tcp_ip_channel.c
```diff
@@ -142,25 +140,25 @@ TaggedMessage *TcpIpChannel_receive(NetworkChannel *untyped_self) {
...
+  // configure select to check if data is available (non blocking)
+  fd_set read_fds;
+  struct timeval timeout;
+  FD_ZERO(&read_fds);
+  FD_SET(socket, &read_fds);
+  timeout.tv_sec = 0;
+  timeout.tv_usec = 0;
+
+  // check if data is available
+  int select_result = select(socket + 1, &read_fds, NULL, NULL, &timeout);
```
LasseRosenow, line 152: I just gonna copy your comment from my previous PR to keep track of it:

@erlingrj wrote: 

Do we actually need to check how much data is there before we read? Since this is running in a separate thread, we want it to block on recv and return with as much data is available. What is gained from checking how much data is there before the read?

(I know that this was there before your edits, and perhaps @tanneberger can answer better?)

erlingrj:
Is this still a Draft or should we consider it for merging, @LasseRosenow? I like this approach since I can use it for Zephyr also

LasseRosenow:
> Is this still a Draft or should we consider it for merging, @LasseRosenow? I like this approach since I can use it for Zephyr also

I will add more examples to RIOT today and see if I can test if everything still works correctly and undraft it :)

erlingrj:
@LasseRosenow I am merging this because I have some other work based off this branch that I want to merge. I hope this doesn't cause any problems for you

PR #62, opened 2024-10-15
Fix spelling of wait_until_interruptible and make platform/riot.c compile

LasseRosenow:
Partially this is to fix a typo and partially it also fixes the compilation of riot.c

Funnily the type did not exist in riot.c so riot.c did not compile :)

I think fixing the typo in the rest of the runtime is more reasonable to make riot.c inherit the typo


erlingrj:
Great, thanks!

PR #63, opened 2024-10-15
Add RIOT examples

LasseRosenow:
The only relevant changes here are in the "examples" directory.

Probably not the most clean solution, but it works for now.

See my comments for some questions.


examples/riot/external_modules/reactor-uc/Makefile.include
```diff
@@ -0,0 +1,9 @@
+
+CFLAGS += -DPLATFORM_RIOT=1
+
+# Use an immediate variable to evaluate `MAKEFILE_LIST` now
+USEMODULE_INCLUDES_reactor-uc := $(LAST_MAKEFILEDIR)/../../../../include
+USEMODULE_INCLUDES += $(USEMODULE_INCLUDES_reactor-uc)
+
+USEMODULE_INCLUDES_reactor-uc2 := $(LAST_MAKEFILEDIR)/../../../../external
```
LasseRosenow, line 8: Regarding this thing I have no Idea why I needed to do it like this, but it works for now. I will ask some people who know the riot build system better to improve this in the future ..

examples/riot/src/blinky/Makefile
```diff
@@ -0,0 +1,28 @@
+# name of your application
+APPLICATION = lf-test
+
+# If no BOARD is found in the environment, use this default:
+BOARD ?= native
+
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/../../../../../RIOT
```
LasseRosenow, line 8: I have specified the RIOT folder to be one higher thant the root of this project. So it is easy to just clone and put the RIOT folder next to this project.

Maybe we can improve the ergonomix here by including RIOT as a submodule?
If this is the way to go, I was not yet sure where to put it. Should we put the RIOT submodule just in the examples/riot folder?

erlingrj (changes requested):
Happy that you are tackling this. A question that we should find an answer to right away is how we should integrate with RIOT. Tassilo started on something in his fork of RIOT (https://github.com/tanneberger/RIOT), but we got hold up because I couldn't make the RIOT make-based build system do what I wanted. 

In Tassilos fork, we added reactor-uc as an external package by pointing to a commit of this repo. We could then write a RIOT programming using the reactor-uc package and linking it. Currently we are thinking of LF as the language for writing the applications so we would have LFC code-generating a project that uses reactor-uc. Normally, LFC also copies the entire reactor-uc library into the project. So I am not sure whether it makes sense to have reactor-uc as an external package in RIOT or just part of each project.

What do you think @tanneberger?

examples/riot/src/blinky/Makefile
LasseRosenow: Yes including this project into riot as a package is easy and a common way to do it.
But for testing this project on riot and doing active development on it I find the RIOT package system inkonvenient. It requires you to point to a specific commit ID and we always need to update the RIOT package to new versions.

Thats why I added this example within this project. So I see 2 options: 
- Option 1: Add RIOT as a submodule to the examples/riot folder, which does not mean this project depends on RIOT, but just makes it more easy for new people to run the riot examples in here.
- Option 2: Keep as is and just explain in the readme that one needs to clone RIOT next to this project on their file system to run the examples.

Both ways are totally reasonable for me, but feel free to come up with an even better solution :);)

examples/riot/src/blinky/Makefile
```diff
@@ -0,0 +1,28 @@
...
+# Change this to 0 show compiler invocation lines by default:
+QUIET ?= 1
+
+# Use a peripheral timer for the delay, if available
+FEATURES_OPTIONAL += periph_timer
+
+# External modules
+USEMODULE += reactor-uc
+USEMODULE += ztimer64
+USEMODULE += ztimer64_usec
```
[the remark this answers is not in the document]
LasseRosenow: I agree. I was not sure on how hard reactor-uc's dependency on ztimer and ztimer_usec is. If you think it is always needed then I would move it up to the external_modules makefile

examples/riot/external_modules/reactor-uc/Makefile
```diff
@@ -0,0 +1,4 @@
+include $(RIOTBASE)/Makefile.base
+
+all:
+	$(QQ)"$(MAKE)" -C ../../../../src -f $(CURDIR)/reactor-uc.mk
```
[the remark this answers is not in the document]
LasseRosenow: I personally prefer not to have so much platform specific files in the root folder, but if you think this is fine then sure we can move the make files for riot into the root folder. This can then also be reused in package of reactor-uc that is included into RIOT OS in @tanneberger's fork :)

examples/riot/src/blinky/Makefile
erlingrj: I prefer 2 but I would like to not rely on relative paths, what about requiring an environmental variable, e.g. `RIOT_PATH` or `REACTOR_UC_RIOT_PATH` or something?

examples/riot/src/blinky/Makefile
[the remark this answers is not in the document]
erlingrj: It is a key dependency, it is what we use to implement that `wait_until` function which puts the runtime to sleep until an absolute time-point in the future

examples/riot/src/blinky/Makefile
LasseRosenow: In the marked line you can see that there already is an environment varialble: "RIOTBASE" this is the common variable name we use everywhere in RIOT OS applications. The relative path is just a fallback if that variable is not specified

examples/riot/external_modules/reactor-uc/Makefile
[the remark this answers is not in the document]
LasseRosenow: Okay then I would propose to put RIOT as a submodule into `external/platform/RIOT` for now.

> we need to add nanopb sources to the build. This file can be used by the RIOT build system, or by another Makefile-based build system if the user wants to integrate it with something different.

Yes this seems reasonable I will see to move the non-riot-specific parts to the root folder.

> I was really struggling with getting the RIOT make system to do what I wanted

The same for me :) The RIOT build system can be quite tough and confusing. I myself are still quite confused about how exactly that Makefile.include file exactly works... I hope I can make the reactor-uc external module better soon.

examples/riot/src/blinky/Makefile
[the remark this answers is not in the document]
LasseRosenow: I had some issues including it in the external_module. For some reason the makefile.include file ignores it if i place the USEMODULEs there. But I am confident I can find a fix soon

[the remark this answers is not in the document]
LasseRosenow: This is fixed now. I these dependencies to the `Makefile.dep` file of the `reactor-uc` `external_module`

LasseRosenow:
Okay I reverted the RIOT submodule :)

LasseRosenow:
I set this to draft because even though the threads work now, there are still some compilation problems with riot because of stricter rules. I will see to disable them somehow in make.

examples/riot/blinky/main.c
```diff
@@ -0,0 +1,51 @@
+#include "board.h"
+#include "reactor-uc/reactor-uc.h"
+
+typedef struct {
```
tanneberger, line 4: if you rebase master you can use the new macro shorthands.

LasseRosenow: Done

LasseRosenow:
@erlingrj I think this should be ready to merge now. Feel free to comment if you find remaining problems with this PR :)

erlingrj (changes requested):
Great! I am able to run the blinky example with the `native` "board". However, I would prefer not to add those TcpIpChannel tests as examples because it duplicates everything and making changes to TcpIpChannel will require changes all over.

In fact, I would like to see TcpIpChannel examples moved to the unit-test folder instead.

So I am happy to merge this if we can remove the TcpIpChannel stuff from RIOT examples

LasseRosenow:
Okay I have rebased with main and fixed the issues.
I have removed the tcp test examples and added a basic hello example just like zephyr has.

src/platform/posix/tcp_ip_channel.c
```diff
@@ -217,10 +217,21 @@ void TcpIpChannel_register_callback(NetworkChannel *untyped_self,
   if (pthread_attr_init(&self->receive_thread_attr) < 0) {
     throw("pthread_attr_init failed");
   }
+/* TODO: RIOT posix-wrappers don't have pthread_attr_setstack yet */
+#ifdef __USE_XOPEN2K
```
LasseRosenow, line 221: I needed to do these changed to the tcp_ip_channel because the RIOT POSIX wrappers don't support `pthread_attr_setstack`. I will try to fix that on RIOT ASAP but for now I added this `define` to support both cases.
It is needed to do it like this because `pthread_attr_setstackaddr` is deprecated. So we only want to use it if not avoidable.

erlingrj: OK, we should figure it out, the reason I added it was because on Zephyr you have to do it that way...

erlingrj:
@LasseRosenow thanks for this, great work. I am merging it and I will resolve the failing tests in #71

PR #65, opened 2024-10-15
Dont require upstream port when constructing a Connection

erlingrj:
This addresses #46 and removes the need to pass a pointer to the upstream port when creating a Connection object


erlingrj (commented):
Self-review looks good

PR #69, opened 2024-10-16
Macro Madness

tanneberger:


erlingrj:
I like this a lot! It makes the idea of using reactor-uc as a standalone library actually feasible!

A few comments:
1. The macro arguments which represent values (e.g. sizes etc) should always be referred to with paranthesis in the macro-body because they might be expressions. You will see this with the other macros I defined.
2. Can we find another name than `CONSTRUCT_X` since we aren't constructing it, but defining the constructor. I would argue that defining the constructor should be part of `DEFINE_X`
3. Connections are still todo. Would be nice with a CONNECT(output, input) macro instead of CONN_REGISTER_UPSTREAM and CONN_REGISTER_DOWNSTREAM.
4. Can we figure out something clever for the reactors? There are some boilerplate fields such as `Trigger * _triggers[numTriggers]`,  `Reaction *_reactions[numReactions` and `Reactor *_children[numChildren]`. It is also the initializing of those fields. Might be very hard, maybe impossible

tanneberger:
@erlingrj 

2.) I guess the reason I left it two different things is that, the constructor needs to know the type of the parent reactor, which is defined after all the reactor components. 

I am happy to merge it but then, we need to empty declaration of the reactor at the top.

Issue #70, opened 2024-10-17
Implement a custom allocator to relax constraint on monotonic scheduling of events


erlingrj:
This was suggested by @edwardalee and @schoeberl. Instead of a circular buffer of event-payloads, we should have an allocator that enables out-of-order scheduling of events. Events would need to contain three fields now:
```
typedef struct {
  tag_t tag;
  trigger_t *trigger;
  void *payload
} event_t;
```  

We would have fixed-size pool of event payloads for each trigger and a simple allocator/free function. This will require some extensive changes throughout the code-base but it will make the overall runtime more flexible and avoid this irritating implicit assumption on monotonic increasing tag on scheduling of actions

PR #71, opened 2024-10-17
Implement event payload allocator

erlingrj:
This PR is quite big and does several things.
1. Introduces an event payload allocator and thus relaxing the constraint on having to schedule actions monotonically. This triggered a bit of a redesign of the Trigger type hierarchy.  #70 
2. Updates LFC to use the new macros.
3. Refactors the macros such that they match how lfc generates code #77 
4. Merges Physical+Logical Action, Delayed+Physical connections and Startup+Shutdown to avoid code duplication #76
5. I commented out the entire federate example on POSIX because this is ported over to macros in #73


src/connection.c
```diff
@@ -115,103 +109,50 @@ void DelayedConnection_cleanup(Trigger *trigger) {
...
+  Trigger *trigger = &_self->super;
   Scheduler *sched = &_self->super.parent->env->scheduler;
-
-  if (value) {
-    self->trigger_data_queue.stage(&self->trigger_data_queue, value);
-  } else {
-    throw("DelayedConnection with untyped value");
+  EventPayloadPool *pool = trigger->payload_pool;
+  if (self->staged_payload_ptr == NULL) {
+    ret = pool->allocate(pool, &self->staged_payload_ptr);
```
tanneberger, line 141: so this implicitly manipulates the pointer :(

include/reactor-uc/connection.h
```diff
@@ -65,19 +39,13 @@ void LogicalConnection_ctor(LogicalConnection *self, Reactor *parent, Port **dow
 struct DelayedConnection {
   Connection super;
   interval_t delay;
-  TriggerDataQueue trigger_data_queue;
+  bool is_physical;
```
tanneberger, line 42: is this for expressing DelayedPhysical Connections or is this redundant because the flag already exists in the Action?

include/reactor-uc/action.h
```diff
@@ -13,52 +13,20 @@ struct Action {
   Trigger super;          // Inherit from Trigger
   interval_t min_offset;  // The minimum offset from the current time that an event can be scheduled on this action.
   interval_t min_spacing; // The minimum spacing between two consecutive events on this action.
-  tag_t previous_event;   // Used to enforce min_spacing
-  TriggerEffects effects; // The reactions triggered by this Action.
-  TriggerSources sources; // The reactions that can write to this Action.
-  TriggerDataQueue trigger_data_queue; // FIFO storage of the data associated with the events scheduled on this action.
+  bool is_physical;
+  tag_t previous_event; // Used to enforce min_spacing
```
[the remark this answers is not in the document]
erlingrj: `min_spacing` is part of the reactor semantics. It can be useful because you bound the "event bandwidth" through a port. If `min_spacing` is 100msec then there will maximum be 10 events/sec scheduled on this action. 

But it is not correctly implemented now since we support arbitrary scheduling. We would need to search the event queue and look for events from the same action happening before and after...

I think we can safely remove it for now. Thanks for comment

src/builtin_triggers.c
```diff
@@ -2,28 +2,18 @@
...
+  assert(self->effects.size > 0);
 
-  if (!effects) {
-    assert(false);
-  } else {
-    for (size_t i = 0; i < effects->size; i++) {
-      validaten(sched->reaction_queue.insert(&sched->reaction_queue, effects->reactions[i]));
-    }
+  for (size_t i = 0; i < self->effects.size; i++) {
+    validaten(sched->reaction_queue.insert(&sched->reaction_queue, self->effects.reactions[i]));
```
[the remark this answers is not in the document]
erlingrj: Yes, I agree. One should not put function calls in macros. Though, `validate` is never optimized away. I will fix it!

include/reactor-uc/connection.h
erlingrj: Here I am merging delayed and physical connection into a single type. DelayedConnection, so it can have a delay and be physical. It does not inherit from Aciton, rather it is a separate kind of trigger

include/reactor-uc/action.h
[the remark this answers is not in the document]
edwardalee: Note that validating real-time properties is not possible (for physical actions at least) without the `min_spacing` parameter.

include/reactor-uc/action.h
```diff
@@ -10,55 +10,21 @@ typedef struct LogicalAction LogicalAction;
...
-  Trigger super;          // Inherit from Trigger
-  interval_t min_offset;  // The minimum offset from the current time that an event can be scheduled on this action.
-  interval_t min_spacing; // The minimum spacing between two consecutive events on this action.
-  tag_t previous_event;   // Used to enforce min_spacing
-  TriggerEffects effects; // The reactions triggered by this Action.
-  TriggerSources sources; // The reactions that can write to this Action.
-  TriggerDataQueue trigger_data_queue; // FIFO storage of the data associated with the events scheduled on this action.
+  Trigger super;         // Inherit from Trigger
+  interval_t min_offset; // The minimum offset from the current time that an event can be scheduled on this action.
+  bool is_physical;
```
LasseRosenow, line 15: I personally prefer enums for this for better code readability in cases like:

```c
if (action.type == ACTION_LOGICAL) {}
```

vs.

```c
if (!action.is_physical()) {}
```

But I don't have a strong opinion on this. I also think it is okay as is.

include/reactor-uc/action.h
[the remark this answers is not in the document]
erlingrj: Yes, I agree. It is useful, I will put it on the backlog to implement later.

src/connection.c
erlingrj: I am not sure what you mean by "implicitly". But for DelayedConnections we must buffer the proposed event until the end of the tag and we can send it out. This is done by having the field `staged_payload_ptr` point to this payload. It can be overwritten by other calls to `lf_set` and at the end of the tag it is sent and `staged_payload_ptr` is reset to point to NULL.

include/reactor-uc/connection.h
tanneberger: How to we express a normal PhysicalConnection then?

This call `is_physical=true` and `delay = 0` ?

erlingrj: Yes, or rather, delay=NEVER, delay=0 means a microstep delay

erlingrj: We should make a separate macro for physical connections also

Issue #79, opened 2024-10-20
Support federated agreement on start tag


erlingrj:
We need a distributed way of agreeing on what the start tag of a federation should be.

We can solve this the hard way using some distributed consensus algorithm. Or, we can solve it the easy way by making the user pick the "leader" and then have the leader select the start tag and distribute it to all others.

PR #81, opened 2024-10-21
Distribute the start tag within the federation

erlingrj:
This PR addresses #79 with following changes:
- Introduce a protobuf message hierarchy with TaggedMessage and various Signals (StartTagSignal being one of them)
- Assumes a very simple setup where one federate must be chosen as the `leader` and only it will send out start tags to others. This will of course not work if several federates think they are the leader
- Any federate that is not the leader will block indefintely waiting for receiving the start_time
- When start_time is received for this first time by a federate it will be forwarded on all other connections.


erlingrj (commented):
I think it looks OK. @LasseRosenow and @tanneberger, would you please take a look? Summarized the changes are:
- Federates are now either leader or not, this defines how the start_tag is obtained.
- I add another level of hierarchy in the messages. I call it FederateMessage which contains TaggedMessage and StartTagSignal currently
- We must now defer scheduling timers and startups until we have acquired the start tag.
- Some pico fixes

tanneberger (approved):

PR #82, opened 2024-10-21
Make Period and Offset configurable at Construction

### LINKED ISSUE #77: Macro system needs some adaptations to work with LFC
A few modifications are needed to work with LFC.
1. We must split out DEFINE_TIMER_STRUCT and DEFINE_TIMER_CTOR since first goes in a header file and second goes in a source file.
2. We must expose (offset, period) and (min_delay, min_spacing) as ctor arguments, not hardcoded. This is to support setting those based on reactor parameters

tanneberger:
- also split `DEFINE_TIMER` macro into `DEFINE_TIMER_STRUCT` and `DEFINE_TIMER_CTOR`

closes #77

### DIFF
--- a/include/reactor-uc/macros.h
+++ b/include/reactor-uc/macros.h
@@ -122,14 +122,15 @@ typedef struct Input Input;
 
 typedef struct Timer Timer;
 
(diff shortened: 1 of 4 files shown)


erlingrj:
Would you rather comment on #71 where I do this change and also for the other macros

PR #91, opened 2024-10-23
Migrate (federated | delayed) connection.is_physical to ENUM

LasseRosenow:
Similar to my feedback on the Action.is_physical I also updated the DelayedConnection and FederatedConnection to use an ENUM instead of the is_physical boolean.

I am not sure if maybe the Kotlin Compiler needs to be adapted to this and I am not confident in that area yet, so I left it out for now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,8 @@ jobs:
         run: |
           source env.bash
           make format-check
-          make test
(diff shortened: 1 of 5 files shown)


erlingrj:
Great, for now I disable the LF tests because we wont waste time working on the code-generator until the APIs have stabilized

PR #93, opened 2024-10-23
Allocating output connection list at common parent reactor 

tanneberger:


erlingrj (approved):
Great, also ported the zephyr federated test to the new system

PR #94, opened 2024-10-24
Migrate to new NetworkChannel interface

### LINKED ISSUE #83: Define the NetworkChannel interface
Lasse, have you made any progress in this regard? I am somewhat blocked on this because I want to implement start-tag distribution and fix akwardness of startup.

My current propsal is this:
```
struct NetworkChannel {
  size_t dest_channel_id; // So that we can "address" one of several NetworkChannel's at the other end.
  lf_ret_t (*open_connection(NetworkChannel *self);
  lf_ret_t (*is_connected(NetworkChannel *self);
  lf_ret_t (*send_blocking)(NetworkChannel *self, TaggedMessage *message);
  void *(register_receive_callback(...)
};
```

Key design decisions here that we can discuss are:
1. It does not make the difference between server/client. This might be a bad decision and we might reconsider and add a flag to the constructor. 
2. It introduces the notion of a `dest_channel_id`. This is an address into the federate on the other side. It lets you say that this message is destined for a particular NetworkChannel at the other side. It is redundant with TCP/UDP since we can just use ports. We might argue that it is not a good idea to have this in the NetworkChannel interface and instead let the different networking stacks handle it.
3. Only `send_blocking` is blocking. I.e. `open_connection` does not guarantee that a connection is established. Only when `is_connected` returns `LF_OK` can we conclude that a connection is established. This has the advantage that a single-threaded runtime can "open" all his connections and then in the end poll all of them in a while loop until all are connected. Any suggestions for an alternative approach?

### LINKED ISSUE #88: Make TcpIpChannel `connect` call non-blocking
Currently we need the server to call `bind` before the client calls `connect`. According to Lasse, this is simply solved by passing a NON_BLOCKING flag to the connect call

LasseRosenow:
Closes #83 
Closes #88


LasseRosenow:
Okay this should be finished now.

I am not sure why these tests fail. It seems my unit test cannot find the tcp_ip_channel.c file, but locally it compiles and runs on my computer.

The client and server seem to be non blocking now I added tests for this also.
The way the test works is a bit rough for now: If the tcp_ip_channel blocks even though not allowed, the test will probably never finish. I am not sure if the CI has timeouts for that.

But as long as everything finishes running at least we know it is non-blocking. :) 🤓

test/unit/mock.c
```diff
@@ -1,3 +1,3 @@
 #include "unity.h"
-void setUp(void) {}
-void tearDown(void) {}
\ No newline at end of file
+// void setUp(void) {}
```
LasseRosenow, line 2: I am not sure. It seems there is something wrong with the cmake configuration.
I am using these functions in the new `network_channels_test.c` and the compiler was complaining about multiple implementations for these functions.

LasseRosenow: By commenting out these implementations in mock.c it compiled.

src/platform/posix/tcp_ip_channel.c
```diff
@@ -250,40 +289,27 @@ void TcpIpChannel_ctor(TcpIpChannel *self, const char *host, unsigned short port
...
-    if (pthread_cancel(self->receive_thread) != 0) {
-      LF_ERR(NET, "Error canceling receive thread");
-    }
-
-    if (pthread_join(self->receive_thread, NULL) != 0) {
-      LF_ERR(NET, "Error joining receive thread");
-    }
-  }
-  self->super.close((NetworkChannel *)self);
+  self->receive_thread = 0;
```
LasseRosenow, line 314: This is needed so that the `free` function does not cleanup threads that don't exist anymore if you clean and recreate the same tcpchannel. (I noticed this when running my new unit test)

LasseRosenow: Because sometimes we recreate the tcp_ip_channel but without a new thread so the thread is still closed in reality

src/platform/posix/tcp_ip_channel.c
```diff
@@ -220,7 +232,7 @@ void TcpIpChannel_register_callback(NetworkChannel *untyped_self,
     throw("pthread_attr_init failed");
   }
 /* TODO: RIOT posix-wrappers don't have pthread_attr_setstack yet */
-#ifdef PLATFORM_RIOT
+#if defined(PLATFORM_RIOT) && !defined(__USE_XOPEN2K)
```
LasseRosenow, line 235: When compiling for RIOT native BOARD we otherwise get a deprecated warning. Native does support the new setstack function

src/platform/posix/tcp_ip_channel.c
```diff
@@ -173,11 +184,11 @@ TaggedMessage *TcpIpChannel_receive(NetworkChannel *untyped_self) {
   return &self->output;
 }
 
-void TcpIpChannel_close(NetworkChannel *untyped_self) {
+static void TcpIpChannel_close_connection(NetworkChannel *untyped_self) {
   LF_DEBUG(NET, "Closing TCP/IP Channel");
   TcpIpChannel *self = (TcpIpChannel *)untyped_self;
 
-  if (self->server) {
+  if (self->server && self->client != 0) {
```
LasseRosenow, line 191: This extra check is needed because in my test I sometimes just open a connection only and don't actually create a client socket. Then cleaning a client socket here creates an error.

include/reactor-uc/platform/posix/tcp_ip_channel.h
```diff
@@ -33,7 +33,6 @@ struct TcpIpChannel {
 
   fd_set set;
   bool server;
-  bool blocking;
```
LasseRosenow, line 36: This variable is unused

LasseRosenow:
Okay I added some comments to explain some important code parts

test/unit/mock.c
erlingrj: Hm, on my end I get undefined reference errors when these functions are not there...

erlingrj: Ah, I just defined them that way because we are not using them for any other test

LasseRosenow:
About the test name:

I named it network channel test because except the setup function the test is implementation independent.

In theory once we have more channels we could run these tests all in one place and share the code.

We only need to call the correct constructors accordingly in setup

But it is okay to call it tcp for now. Maybe we can revisit once we have more channels

src/platform/posix/tcp_ip_channel.c
```diff
@@ -132,7 +144,7 @@ lf_ret_t TcpIpChannel_send(NetworkChannel *untyped_self, TaggedMessage *message)
   return LF_OK;
 }
 
-TaggedMessage *TcpIpChannel_receive(NetworkChannel *untyped_self) {
+static TaggedMessage *TcpIpChannel_receive(NetworkChannel *untyped_self) {
```
erlingrj, line 147: I like static functions, but I like const pointers better, we should start using them also

erlingrj:
This looks great to me. A good first step towards stabilizing the NetworkChannel!

src/platform/posix/tcp_ip_channel.c
LasseRosenow: I agree ;)

PR #98, opened 2024-10-25
Shutdown

erlingrj:
- Add `request_shutdown` that can be called from the environment
- Improve a little how shutdown is done in the Scheduler
- Catch ctrl-c from Platform_posix and call `lf_exit` if it is defined


erlingrj (commented):
Looks good I think

tanneberger (approved):

PR #100, opened 2024-10-25
Build Zephyr examples in CI

erlingrj:


erlingrj:
Failure in pico is fixed in the distribute start tag PR

PR #104, opened 2024-10-28
Build RIOT examples in ci

### LINKED ISSUE #101: Add CI action for building the RIOT example
@LasseRosenow is this something you could look at? You could reuse the structure used for Zephyr and Pico. I added caching to speed things up

LasseRosenow:
Closes #101

### DIFF
--- a/.github/workflows/riot.yml
+++ b/.github/workflows/riot.yml
@@ -7,6 +7,11 @@ jobs:
   ci:
     name: Build RIOT examples
     runs-on: ubuntu-latest
+    container:
+      image: riotbuild:latest
+      env:
+        RIOT_VERSION: "2024.07"
+        RIOTBASE: "/opt/RIOT-2024.07"
     env:
       RIOT_VERSION: "2024.07"
       RIOTBASE: "/opt/RIOT-2024.07"
--- a/.github/workflows/riot.yml
+++ b/.github/workflows/riot.yml
@@ -1,0 +1,25 @@
+name: RIOT examples
+
+on:
+  pull_request:
+
+jobs:
+  ci:
+    name: Build RIOT examples
+    runs-on: ubuntu-latest
+    env:
+      RIOT_VERSION: "2024.07"
+      RIOTBASE: "/opt/RIOT-2024.07"
+    steps:
+      - name: Checkout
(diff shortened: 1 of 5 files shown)


LasseRosenow:
This should be done now and is ready for review

.github/workflows/riot.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  ci:
+    name: Build RIOT examples
+    runs-on: ubuntu-latest
+    container:
+      image: riot/riotbuild:latest
+    env:
+      RIOT_VERSION: "2024.07"
+      RIOTBASE: "/opt/riot-2024.07"
```
LasseRosenow, line 14: This variable is needed by make to know where the RIOT sources can be found.
And since it is already there I reused it in the action to specify the path etc. to make it more durable

examples/riot/buildAll.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+# List of folders
+FOLDERS=("blinky" "hello")
+
+# List of boards
+BOARDS=("native" "nucleo-f429zi")
```
LasseRosenow, line 9: I allow to build multiple boards so that I can at least test `native` and an example board.
The `native` target runs `RIOT` on your linux system and usually can behave quite different to a concrete board.
So I think it is important to always also test for `native`.

erlingrj (approved):
Looks great. lets get it merged :rocket:

LasseRosenow:
Yeah actually the RIOT compiler is quite strict. I have already disabled a few rules to make it run.
I think for now I could disable this rule also to make the ci run for now, but we should fix it in another PR

erlingrj:
Yeah, it appears that GNU C allows it, but not ISO C. We are currently using several GCC extensions so technically we arent ISO C compatible. That could be an issue

LasseRosenow:
Nice the test now runs successfully :)
I opened #105 to not forget about the compiler issues, or to at least move the discussion there if we want to be ISO-C compatible or not etc.

LasseRosenow:
Sorry I accidentally merged using the wrong button and moved all the commits to the main branch.
I fixed it now

PR #106, opened 2024-10-29
Improve handling of non-blocking try-connect

erlingrj:
This PR was triggered by not getting the Zephyr federated example to work on our boards anymore. It does several things:
- Improve the error handling in the non-blocking try-connect. Basically checks for EINPROGRESS and then does `select` to see that socket is writable before checking error message. Possibly closing socket and reopening if it failed.
- Reject tagged messages and actions before start tag is resolved
- Move serialization code into serialization.c/h


src/platform/posix/tcp_ip_channel.c
```diff
@@ -52,32 +76,73 @@ static lf_ret_t TcpIpChannel_open_connection(NetworkChannel *untyped_self) {
...
+      return LF_ERR;
+    }
+  }
+}
 
-    new_socket = accept(self->fd, (struct sockaddr *)&address, &addrlen);
-    if (new_socket >= 0) {
-      self->client = new_socket;
-      FD_SET(new_socket, &self->set);
+static lf_ret_t check_if_socket_is_writable(int fd) {
```
LasseRosenow, line 103: This and other function should maybe start with `TcpIpChannel_`. At least this is our current code style as I can remember?
I personally prefer to have static functions start with an underscore, it makes it more clear, which functions are actually public for me and in the IDE, but better have it consistent through the project

src/platform/posix/tcp_ip_channel.c
```diff
@@ -89,11 +154,53 @@ static lf_ret_t TcpIpChannel_try_connect(NetworkChannel *untyped_self) {
...
     }
   }
+  return LF_ERR; // Should never reach here
+}
 
-  return LF_OK;
+static lf_ret_t TcpIpChannel_try_connect(NetworkChannel *untyped_self) {
+  TcpIpChannel *self = (TcpIpChannel *)untyped_self;
+  if (self->server) {
+    return TcpIpChannel_try_connect_server(untyped_self);
```
LasseRosenow, line 200: Since the `try_connect_server` and `try_connect_client` functions are only called within this function it could make sens to make them inline functions to allow for compiler optimizations?

LasseRosenow: The same could also be said for all the other helper functions maybe

src/platform/posix/tcp_ip_channel.c
```diff
@@ -280,19 +403,6 @@ static void TcpIpChannel_free(NetworkChannel *untyped_self) {
 }
 
 void TcpIpChannel_ctor(TcpIpChannel *self, const char *host, unsigned short port, int protocol_family, bool server) {
-  FD_ZERO(&self->set);
```
LasseRosenow, line 283: Moving this to the reset function is very nice

src/platform/posix/tcp_ip_channel.c
```diff
@@ -89,11 +154,53 @@ static lf_ret_t TcpIpChannel_try_connect(NetworkChannel *untyped_self) {
...
-      return LF_COULD_NOT_CONNECT;
+      if (errno == EINPROGRESS) {
+        self->client_connect_in_progress = true;
+        LF_DEBUG(NET, "Connection in progress!");
+        return LF_TRY_AGAIN;
+      } else {
+        LF_ERR(NET, "Connect failed errno=%d", errno);
+        self->client_connect_in_progress = false;
+        TcpIpChannel_reset_socket(self);
+        return LF_TRY_AGAIN;
```
LasseRosenow, line 165: At first it was a little bit confusing for me that in progress and failed both return `TRY_AGAIN`, but I guess it does somewhat make sense from the `NetworkChannel` perspective.
In both cases we need to try again.

But I see one argument why maybe a differentiation could make sense here.

In the Zephyr example you sleep for some time if you get a `TRY_AGAIN`, but this only makes sense for the `EINPROGRESS` case from what I can see.

In case of a `reset_socket` we could just immediately try again.

So maybe return `LF_IN_PROGRESS` For `EINPROGRESS` and only wait for that case?

LasseRosenow: And then also maybe we could be more transparent and instead of `LF_TRY_AGAIN` return a `LF_CONNECTION_FAILED`? The user of the API can then decide if they want to try again or not I guess?
But I am also okay with keeping `TRY_AGAIN`. It has the advantage of telling the user what to do ;)

LasseRosenow: Another way would be `LF_IN_PROGRESS_TRY_AGAIN` and `LF_FAILED_TRY_AGAIN`

LasseRosenow:
Looking good to me, I left some small comments :)
And nice to have more LF_DEBUG etc. That is gonna help a lot!

src/platform/posix/tcp_ip_channel.c
erlingrj: Hmm. The `TcpIpChannel_` prefix was initially meant for naming functions that had many implementations for different "subclasses". They are meant for implementing the corresponding functions in the public API (which is the function pointers on the struct). For purely static functions it might be different. But it might get messy if we dont pick a single convetion. I will follow your suggestion for now

src/platform/posix/tcp_ip_channel.c
erlingrj: I went for `LF_IN_PROGRESS` and `LF_TRY_AGAIN`, on Zephyr we do a short sleep before trying again.

PR #107, opened 2024-10-29
Actions without Types

tanneberger:


include/reactor-uc/macros.h
```diff
@@ -28,12 +28,27 @@
...
  */
-#define lf_schedule(action, val, offset)                                                                               \
+
+#define lf_schedule_with_val(action, offset, val)                                                                      \
   do {                                                                                                                 \
     __typeof__(val) __val = (val);                                                                                     \
-    (action)->super.schedule(&(action)->super, (offset), (const void *)&__val);                                        \
+    lf_ret_t ret = (action)->super.schedule(&(action)->super, (offset), (const void *)&__val);                         \
+    if (ret == LF_FATAL) {                                                                                             \
+      throw("no memory in action!");                                                                                   \
```
erlingrj, line 37: I would say that it is not a fatal error that schedule fails. It might happen if this is a physical action and the runtime is lagging too much behind. But can we return a value from this `lf_schedule_with_val` macro? There might be some GCC extension allowing it.

include/reactor-uc/macros.h
```diff
@@ -188,18 +210,34 @@
...
   } ActionName;
 
-#define DEFINE_ACTION_CTOR_FIXED(ActionName, ActionType, EffectSize, SourceSize, BufferType, BufferSize, MinDelay)     \
+#define GET_ARG9(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, ...) arg9
+#define GET_ARG8(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, ...) arg8
+#define GET_ARG7(arg1, arg2, arg3, arg4, arg5, arg6, arg7, ...) arg7
+#define GET_ARG6(arg1, arg2, arg3, arg4, arg5, arg6, ...) arg6
+
+#define DEFINE_ACTION_STRUCT_CHOOSER(...)                                                                              \
+  GET_ARG7(__VA_ARGS__, DEFINE_ACTION_STRUCT_WITH_VALUE, NULL, DEFINE_ACTION_STRUCT_WITHOUT_VALUE)
```
erlingrj, line 219: Would be great with a short descripbtion of why it works (and its assumptions on the parmeters to DEFINE_ACTION_STRUCT.

src/action.c
```diff
@@ -30,24 +30,33 @@ lf_ret_t Action_schedule(Action *self, interval_t offset, const void *value) {
...
-  validate(value);
 
   env->enter_critical_section(env);
 
+  // user tried to schedule a action that does not have any value
+  if (self->super.payload_pool->capacity == 0 && value != NULL) {
+    env->leave_critical_section(env);
+    LF_ERR(TRIG, "Scheduling an value, that doesn't have value!");
+    sched->do_shutdown(sched, sched->current_tag);
+    return LF_FATAL;
```
erlingrj, line 41: Not a fatal error

erlingrj (changes requested):
A good start

include/reactor-uc/macros.h
tanneberger: `LF_FATAL` is only returned if you schedule an action, that has no value with a value.

erlingrj: OK, then I will update the error message to be a little more descriptive >P

src/action.c
```diff
@@ -34,24 +35,38 @@ lf_ret_t Action_schedule(Action *self, interval_t offset, const void *value) {
...
   }
 
-  ret = self->super.payload_pool->allocate(self->super.payload_pool, &payload);
-  if (ret != LF_OK) {
-    return ret;
+  if (self->super.payload_pool->capacity == 0 && value != NULL) {
+    // user tried to schedule a action that does not have any value
+    env->leave_critical_section(env);
+    LF_ERR(TRIG, "Scheduling an value, that doesn't have value!");
+    sched->do_shutdown(sched, sched->current_tag);
```
erlingrj, line 53: I would say that it is the callers job of calling `do_shutdown`, should it instead be done in the `lf_schedule` macro?  ideally it would be in the `throw` macro, but it then needs a pointer to the environment...

erlingrj (changes requested):
This looks mostly good now. Only a few nitpicks...

PR #111, opened 2024-11-01
Support scheduling multiple events with the same tag onto action

erlingrj:
We solve it "lazily" add the beginning of a tag by just overwriting the `value` field of the action and not enqueue the reactions if they already have been.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -19,10 +19,10 @@ void Action_prepare(Trigger *self, Event *event) {
   if (self->is_present) {
     LF_WARN(TRIG, "Action %p is already present at this tag. Its value was overwritten", self);
   } else {
-  sched->register_for_cleanup(sched, self);
-  for (size_t i = 0; i < act->effects.size; i++) {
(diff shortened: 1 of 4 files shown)


src/action.c
```diff
@@ -14,14 +14,18 @@ void Action_prepare(Trigger *self, Event *event) {
...
-  sched->register_for_cleanup(sched, self);
-
-  for (size_t i = 0; i < act->effects.size; i++) {
-    validaten(sched->reaction_queue.insert(&sched->reaction_queue, act->effects.reactions[i]));
+  if (self->is_present) {
+    LF_WARN(TRIG, "Action %p is already present at this tag. Its value was overwritten", self);
+  } else {
+    sched->register_for_cleanup(sched, self);
+    for (size_t i = 0; i < act->effects.size; i++) {
+      validaten(sched->reaction_queue.insert(&sched->reaction_queue, act->effects.reactions[i]));
```
tanneberger, line 24: `sed s/validaten/validate`

tanneberger (changes requested):

src/action.c
erlingrj: done

PR #113, opened 2024-11-02
Add DECLARE_CTOR macros

erlingrj:
We need to declare the ctors in header files and define them in source files


erlingrj:
This is actually not necessary

PR #117, opened 2024-11-08
Remove `dest_channel_id` from NetworkChannel

LasseRosenow:
I think it is more reasonable to not duplicate this value if we anyways for example have the ports or file descriptors stored in the concrete channel implementation structs.

### DIFF
--- a/src/platform/posix/tcp_ip_channel.c
+++ b/src/platform/posix/tcp_ip_channel.c
@@ -469,7 +469,7 @@ static void TcpIpChannel_free(NetworkChannel *untyped_self) {
   self->super.close_connection((NetworkChannel *)self);
 }
 
-NetworkChannelState TcpIpChannel_get_state(NetworkChannel *untyped_self) {
+static NetworkChannelState TcpIpChannel_get_connection_state(NetworkChannel *untyped_self) {
   TcpIpChannel *self = (TcpIpChannel *)untyped_self;
   return self->state;
 }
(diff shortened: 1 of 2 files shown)


src/platform/posix/tcp_ip_channel.c
```diff
@@ -486,6 +496,8 @@ void TcpIpChannel_ctor(TcpIpChannel *self, const char *host, unsigned short port
   self->fd = 0;
   self->state = NETWORK_CHANNEL_STATE_UNINITIALIZED;
 
+  self->super.get_dest_channel_id = TcpIpChannel_get_dest_channel_id;
+  self->super.get_connection_state = TcpIpChannel_get_connection_state;
```
LasseRosenow, line 500: It seems the `connection_state` mapping was missing so I secretly added it here also

src/platform/posix/tcp_ip_channel.c
```diff
@@ -469,7 +469,17 @@ static void TcpIpChannel_free(NetworkChannel *untyped_self) {
...
-NetworkChannelState TcpIpChannel_get_state(NetworkChannel *untyped_self) {
+static uint32_t TcpIpChannel_get_dest_channel_id(NetworkChannel *untyped_self) {
+  TcpIpChannel *self = (TcpIpChannel *)untyped_self;
+
+  if (self->server) {
+    return self->client;
+  } else {
+    return self->fd;
+  }
+}
```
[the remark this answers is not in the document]
erlingrj: This depends on how `get_dest_channel_id` is to be used by the runtime. In fact, I think it might not be used by the runtime at all, only by the lower levels of the NetworkChannel implementation. This suggests that it might not be the right decision to have this function as part of the generic NetworkChannel API, rather it some bookkeeping that must be done in the different concrete implementations?

src/platform/posix/tcp_ip_channel.c
erlingrj: Ops! Thanks!

LasseRosenow:
Okay I rebased this and removed the `conn_channel_id` from the `NetworkChannel`

The other small change is that I renamed the `get_connection_state` function in the `TcpIpChannel` to match the name of the `NetworkChannel` interface.

erlingrj (approved):
Great!

PR #118, opened 2024-11-09
More macro system restructuring

erlingrj:
This PR makes another iteration on the macro system. It replaces even more boilerplate and makes the reactor setup easier to understand. Changes summarized:
- Add `REACTION_INSTANCE()`++ that creates the instance on the self struct of the parent reactor
- Add `INITIALIZE_REACTION()`++ that is to be used in the parent reactor constructor to initialize (call constructor++) the instance
- All the macros now takes name of the parent as the first argument. This avoids naming collisions for bigger programs. So you write `DEFINE_TIMER_STRUCT(MyReactor, timerName, 1)` and the typename will be `MyReactor_timerName`
- Reactions are now named
- When registering sources and effects we now pass the _name_ of the reactor element, not a pointer to it. So you do: `TIMER_REGISTER_EFFECT(myTimer, myReaction)` not needing the `self->myTimer`  etc.

I did remove the fancy macro for specifying an action without a value. The reason is that since it is not user-facing (it will be code generated) I think the added complexity is a bigger drawback than convenience. I kept it for `lf_schedule` which is user-facing and there I think the benefits are bigger.


erlingrj:
Finally have ported all the unit tests over to the "new" macro system. Only missing the examples

erlingrj:
@tanneberger :pray:

PR #120, opened 2024-11-13
Add RIOT CoAP Channel

LasseRosenow:
CoAP channel for the RIOT platform.

All channels share the same server and delegate the message to the correct Channel based on the other federates IP-Address.


examples/riot/coap_channel/Makefile
```diff
@@ -0,0 +1,28 @@
+# name of your application
+APPLICATION = lf-test
```
erlingrj, line 2: Should we call it something more descriptive?

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+}
+
+static void CoapChannel_free(NetworkChannel *untyped_self) {
+  LF_DEBUG(NET, "CoapChannel: Free");
+  CoapChannel *self = (CoapChannel *)untyped_self;
+
+  // TODO
+}
+
+void CoapChannel_ctor(CoapChannel *self, Environment *env, const char *remote_host) {
```
erlingrj, line 275: So this is now hard-coded COAP over UDP? We might want to create a unique channel for each physical layer that all shares some common COAP implementation. `UdpCoapChannel`, `BleCoapChannel`, `ZigbeeCoapChannel`.

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+  // Concrete fields
+  self->receive_callback = NULL;
+  self->federated_connection = NULL;
+  self->state = NETWORK_CHANNEL_STATE_UNINITIALIZED;
+
+  // Convert host to udp socket
+  sock_udp_name2ep(&self->remote, remote_host);
+  if (self->remote.port == 0) {
+    self->remote.port = CONFIG_GCOAP_PORT;
+  }
```
erlingrj, line 304: If user doesnt specify a port of the remote, then how can we address the particular CoapChannel at the remote (he might have several?)

include/reactor-uc/network_channel.h
```diff
@@ -16,6 +17,7 @@ typedef enum {
   NETWORK_CHANNEL_STATE_CONNECTED,
   NETWORK_CHANNEL_STATE_DISCONNECTED,
   NETWORK_CHANNEL_STATE_LOST_CONNECTION,
+  NETWORK_CHANNEL_STATE_CLOSED,
```
LasseRosenow, line 20: This option was missing.
I needed to know if the `CoapChannel` was willingly `closed` on the other federate.
If the channel is closed then a connection to it will always fail.
This is different to `lost_connection` or `disconnected` as the other channel in this case still wants to connect.

For example:

Following states allow the establishment of a new connection:
- NETWORK_CHANNEL_STATE_OPEN
- NETWORK_CHANNEL_STATE_CONNECTION_IN_PROGRESS
- NETWORK_CHANNEL_STATE_CONNECTED
- NETWORK_CHANNEL_STATE_DISCONNECTED

In contrast these states do not allow the establishment of a connection:
- NETWORK_CHANNEL_STATE_UNINITIALIZED
- NETWORK_CHANNEL_STATE_CLOSED

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+  self->state = NETWORK_CHANNEL_STATE_OPEN;
+}
+
+static void _client_try_connect_callback(const gcoap_request_memo_t *memo, coap_pkt_t *pdu,
+                                         const sock_udp_ep_t *remote) {
+  LF_DEBUG(NET, "CoapChannel: Try connect callback");
+  CoapChannel *self = _get_coap_channel_by_remote(remote);
+
+  if (memo->state == GCOAP_MEMO_TIMEOUT || coap_get_code_class(pdu) != COAP_CLASS_SUCCESS) {
+    self->state = NETWORK_CHANNEL_STATE_DISCONNECTED;
```
erlingrj, line 184: Do we need to acquire a lock before modifying this variable? THis might be called from a different thread?

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+  gcoap_resp_init(pdu, buf, len, COAP_CODE_CONTENT);
+  coap_opt_add_format(pdu, COAP_FORMAT_TEXT);
+  size_t resp_len = coap_opt_finish(pdu, COAP_OPT_FINISH_PAYLOAD);
+
+  // TODO
+
+  return true;
+}
+
+static const coap_resource_t _resources[] = {
```
erlingrj, line 156: This is how we register callbacks for certain incoming messages?

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+
+  // Disconnect successful
+  return true;
+}
+
+static ssize_t _server_message_handler(coap_pkt_t *pdu, uint8_t *buf, size_t len, coap_request_ctx_t *ctx) {
+  LF_DEBUG(NET, "CoapChannel: Server message handler");
+  CoapChannel *self = _get_coap_channel_by_remote(ctx->remote);
+
+  gcoap_resp_init(pdu, buf, len, COAP_CODE_CONTENT);
```
erlingrj, line 134: Do we always need to send a response in COAP

examples/riot/coap_channel/coap_channel.c
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+
+static ssize_t _server_disconnect_handler(coap_pkt_t *pdu, uint8_t *buf, size_t len, coap_request_ctx_t *ctx) {
+  LF_DEBUG(NET, "CoapChannel: Server disconnect handler");
+  CoapChannel *self = _get_coap_channel_by_remote(ctx->remote);
+
+  gcoap_resp_init(pdu, buf, len, COAP_CODE_CONTENT);
+  coap_opt_add_format(pdu, COAP_FORMAT_TEXT);
+  size_t resp_len = coap_opt_finish(pdu, COAP_OPT_FINISH_PAYLOAD);
```
erlingrj: What is this `resp` thing? Are we sending a message somewhere?

examples/riot/coap_channel/main.c
erlingrj, line 1: I am now merging a big update to the macro system, this file will have to be updated to reflect this.

erlingrj (commented):
Looking good, just some clarifying questions. I think we have to start worrying about race conditions now that we introduce callbacks...

examples/riot/coap_channel/Makefile
LasseRosenow: Of course the makefile is super WIP right now I just make it compile and take care of that later :)

examples/riot/coap_channel/coap_channel.c
LasseRosenow: For now yes, I need to see how to abstract the protocol (connection, endpoint logic) from the underlying layers when I explore them deeper as a next step.
RIOT is in a special place here as it likes to run IP over almost everything :D

examples/riot/coap_channel/coap_channel.c
LasseRosenow: The CoAP channel uses the default CoAP port. This decision was made after it turned out to be very hard to have multiple CoAP servers on one device using different ports. The CoAP library in RIOT sets the server port based on that define :( 

But after I thought about this more deeply I think this is actually fine.
We have just one CoAP server yes, but we have multiple channels. When the server gets a request it passes it on to the correct channel based on the IP address of the other federate that the request originates from.

examples/riot/coap_channel/coap_channel.c
LasseRosenow: CoAP is very similar to HTTP in that sense so it is usually expected that you need to respond. I am not 100% sure though.

LasseRosenow: But you don't need to send content of course. It is fine to just return a 204 no-content for example

examples/riot/coap_channel/coap_channel.c
LasseRosenow: This is how we register callbacks for CoAP endpoints just like HTTP endpoints.
If we want to have endpoints per message type is up for debate if it makes sense, but it is definitely doable.

examples/riot/coap_channel/coap_channel.c
LasseRosenow: I need to investigate this I am not sure about the threading situation in gcoap right now

examples/riot/coap_channel/main.c
LasseRosenow: That is no problem, it is almost totally copied from posix, thanks for the heads up!

examples/riot/coap_channel/main.c
```diff
@@ -0,0 +1,185 @@
...
+
+// TODO This macro does not work because I need env here.
+// FEDERATED_CONNECTION_BUNDLE_CTOR_SIGNATURE(Sender, Receiver) {
+//   FEDERATED_CONNECTION_BUNDLE_CTOR_PREAMBLE();
+//   CoapChannel_ctor(&self->channel, env, REMOTE_HOST);
+
+//   FEDERATED_CONNECTION_BUNDLE_CALL_CTOR();
+
+//   INITIALIZE_FEDERATED_OUTPUT_CONNECTION(Sender, out, serialize_msg_t);
+// }
```
LasseRosenow, line 113: @erlingrj With the new macro system even the function name here is defined by a macro, but I need a reference to `Environment` to pass it to the new `CoapChannel`.
As a workaround I have defined the function myself without the macro.

examples/riot/coap_channel/main.c
```diff
@@ -0,0 +1,185 @@
...
+  FEDERATE_BOOKKEEPING_INSTANCES(0, 0, 1, 1);
+  CONTAINED_OUTPUT_CONNECTIONS(sender, out, 1);
+} MainSender;
+
+REACTOR_CTOR_SIGNATURE(MainSender) {
+  FEDERATE_CTOR_PREAMBLE();
+  INITIALIZE_CHILD_REACTOR_WITH_PARAMETERS(Sender, sender, self->_conns_sender_out_out, 1);
+  // TODO this function has more arguments now because I need env
+  // INITIALIZE_FEDERATED_CONNECTION_BUNDLE(Sender, Receiver);
+  Sender_Receiver_Bundle_ctor(&self->Sender_Receiver_bundle, env, &self->super);
```
LasseRosenow, line 139: This is also a workaround because the Macro here also does not allow me to provide an additional parameter `Environment *env`.

examples/riot/coap_channel/main.c
erlingrj: I will fix this

examples/riot/coap_channel/coap_channel.c
erlingrj: After our discussion, and some exploration I think that hardcoding it ontop of UDP makes sense since this seems to be provided for many networking protocols on both Zephyr and RIOT

LasseRosenow:
I have added a test under test/unit/riot/coap_channel_test and restructured this PR to be ready for review so far I would say.

I am really not so sure how to integrate the coap_channel_test with cmake yet, but at least it is possible to run it when using make all in the folder

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -0,0 +1,348 @@
...
+
+  coap_opt_add_format(&pdu, COAP_FORMAT_TEXT);
+  ssize_t len = coap_opt_finish(&pdu, COAP_OPT_FINISH_PAYLOAD);
+
+  // Serialize message into CoAP payload buffer
+  int payload_len = serialize_to_protobuf(message, pdu.payload, pdu.payload_len);
+
+  if (payload_len < 0) {
+    LF_ERR(NET, "CoapUdpIpChannel: Could not encode protobuf");
+    return LF_ERR;
```
erlingrj, line 77:
```suggestion
    return false;
```

erlingrj (changes requested):
This is looking very good Lasse! Great work! In the TcpIpChannel we use the callback `network_channel_state_changed` from the external receive thread if the `recv` returns an error because the sender closed the socket. Is that also necessary here? The reason we need to inform the runtime that it should no longer wait for events on that port (we set the `last_known_tag` to FOREVER). 

I think the handling of changed network channel state is going to need an overhaul in the runtime, it is a little ad hoc.

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -0,0 +1,348 @@
...
+  // Error => return 401 (unauthorized)
+  if (self == NULL) {
+    LF_ERR(NET, "CoapUdpIpChannel: Server disconnect handler: Client has unknown IP address");
+    return gcoap_response(pdu, buf, len, COAP_CODE_UNAUTHORIZED);
+  }
+
+  // Update state because it does not make sense to send data to a closed connection.
+  // Only set state to disconnected if not already in more restrictive state
+  if (self->state != NETWORK_CHANNEL_STATE_UNINITIALIZED && self->state != NETWORK_CHANNEL_STATE_CLOSED) {
+    self->state = NETWORK_CHANNEL_STATE_CLOSED;
```
erlingrj, line 137: Here I think we must call `network_channel_state_updated` so that the runtime knows about it?

erlingrj: I think that function also might need a critical section to avoid race condition

LasseRosenow: >Here I think we must call network_channel_state_updated so that the runtime knows about it?

Should we maybe always call this callback whenever we change state?

erlingrj: Perhaps, I think this whole interaction (runtime <-> network channels) needs some careful thought. What we should avoid is any ambiguity in how information is communicated between them. Currently, for the API calls that deal with setting up the configuration: `open_connection`, `try_connect`, `try_reconnect`, `close_connection`, the state change is implied by the return code. If `LF_IN_PROGRESS` is returned, then the state should be `NETWORK_CHANNEL_STATE_CONNECTION_IN_PROGRESS`. So it is already quite bad that we also have this `network_channel_state_updated` callback which is called sometimes when the state is changed.

Maybe it is better to totally remove that callback, just modify the `state` variable. Then we have to always check the network channel state before we do any blocking wait on input. 

What do you think? I am worried that the interactions with the network channel will be the most complicated problem for our runtime, and that we must keep the interfaces very clear

LasseRosenow: Yes I also found it a little bit duplicated to have these `LF_IN_PROGRESS` return codes and also a state with similar values. But we do need the state and it is useful to have `internally`.

I think because of the confusion it creates maybe ´NOT` expose it to outside of the channel and keep the state internal.

Instead of checking from the runtime each time we should maybe check inside of the channel functions if we can do the operation based on our internal state and if not return an appropriate `LF_ERR` code?

This way the runtime just gets an error code non-blocking of course and can continue just like it checks it itself, but the API is more clear and not duplicated between LF return codes and internal state.

What do you think?

LasseRosenow:
Okay I have now made the unit test run in CI as well.
I have also renamed our CI workflows to make it more clear what they are about and reduced duplication (I think the jobs don't need to contain the word "zephry" or "RIOT" if the wokflow is already called so. :)

src/platform/riot/coap_udp_ip_channel.c
LasseRosenow: @erlingrj Sorry for the ping, just want to make sure, this is not lost in the amount of commits I pushed today while testing the CI 😳

LasseRosenow: PS: I will think about the race condition tomorrow ;)

erlingrj: Hm, the problem is that the "blocking" is actually not performed in the network channel (sorry I was not very accurate). Currently in `Scheduler_federated_acquire_tag` we check the `last_known_tag` of each input port and possibly go to interruptible sleep waiting for the time to elapse. 

In this function, or someplace before, we must check the state of the connection to detect whether the connection has been closed since last time we checked. We must be weary of a race condition here also, this function is executed in a critical section, but we must take care to call `new_async_event` on `Platform` if the connection state changes asynchronously (e.g. receive thread returns with EOF.

If you are interested in proposing something better here, I would gladly review it. We should probably merge this PR and handle this in a separate PR

LasseRosenow: I think maybe we should discuss this in the next meeting on Thursday. I am not sure if I follow you correctly here

erlingrj: Yes, lets discuss it then!

erlingrj (approved):
This is really great work, Lasse! I am very excited to try out the CoapChannel on our end also!

erlingrj:
I think the problem with clang-format is that you have a different version than the CI:
```shell
> clang-format --version
Ubuntu clang-format version 18.1.3 (1ubuntu1)
```

Is it possible for you to get that exact version?

LasseRosenow:
So I installed clang-format-18 and configured this project to use that specific formatter version instead and it works fine now. :)

PR #128, opened 2024-11-27
Remove any platform specific code in code-generated CMakeLists

erlingrj:
This removes the platform specific code-gen CMake. The exception is when the target is  AUTO (not specified) in which case it generates a standalone POSIX project and compiles it directly and puts executable in the bin.

For embedded targets, we get a generic CMakeLists.txt to be included in the build system of the user.


examples/zephyr/hello_lf/CMakeLists.txt
erlingrj, line 1: This is how the user CMakeLists looks, it includes the generated CMakeLists.txt and uses some variables set there, namely: REACTOR_UC_PATH, LF_SOURCES_ABSOLUTE, LF_INCLUDE_DIRS

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
```diff
@@ -17,57 +17,46 @@ class UcCmakeGenerator(private val targetConfig: TargetConfig, private val fileC
...
-
+            generateCmakeEmbedded(sources)
         }
+
+    fun generateCmakeEmbedded(sources: List<Path>) = with(PrependOperator) {
+        """
+            |# This file is generated by LFC. It is meant to be included in
+            |# an existing CMake project. 
+            |
+            |set(LF_SOURCES_RELATIVE
```
erlingrj, line 32: This might be redundant, and we might always want just the absolute paths

PR #140, opened 2024-11-28
Add very simple benchmarking against reactor-c

erlingrj:
This should be merged AFTER #137 

This PR adds the Savina PingPong and a ReactionLatency benchmark, with a C and uC version. Current results on my machine is:

```shell
PingPongUc:
 Best Time:      77.571 msec
Worst Time:      87.062 msec
Median Time:     77.455 msec
PingPongC:
 Best Time:      122.188 msec
Worst Time:      130.564 msec
Median Time:     123.531 msec
ReactionLatencyUc:
 Best latency: 17358. 
Median latency: 56424
Worst latency: 255700
ReactionLatencyC:
 Best latency: 4804
Median latency: 53017
Worst latency: 307132
```

I.e. our runtime has less overhead (PingPong only needs 60-70% of the time needed by reactor-c). When it comes to reaction latency, we are very similar to reactor-c, the Median latency is the most useful number showing that we might have a slightly longer execution path from wakeup to the first reaction. But not significantly. Those last numbers are in nsec btw


tanneberger:
vewwy vewwy cool. 

can you also post the binary sizes and ram usage? of those two examples?

PR #142, opened 2024-11-29
Add RIOT federated CoAP example and fix CoapUdpIp channel implementation

LasseRosenow:
For some reason the `receive` example fails to sleep when `PlatformRiot_wait_for` is called.
I haven't found out why this is the case. But it must be caused somewhere in the runtime. Maybe the thread gets woken up.
With the main.c code from the `sender` it sleepy correctly.

I also fixed some bugs in the CoAP channel implementation.


erlingrj:
Where is `wait_for` called? Is it in the code connecting the federates doing the wait_for between each try_connect? I don't think `wait_for` has really been tested on RIOT yet, but the implementation is straight-forward so its hard to imagine a bug there? You could try inserting a wait_for in a normal program and see that it works?

LasseRosenow:
> Where is `wait_for` called? Is it in the code connecting the federates doing the wait_for between each try_connect? I don't think `wait_for` has really been tested on RIOT yet, but the implementation is straight-forward so its hard to imagine a bug there? You could try inserting a wait_for in a normal program and see that it works?
> 

"Wait for" itself works at least for the sender strangely. I played around with it for quite a bit but haven't found out why it works there but not for the receiver .. I will see further debug it tomorrow

LasseRosenow:
@erlingrj So I found out why the `wait_for` doesn't work here.
The `wait_for` is used within a critical section inside `lf_start` and since interrupts are disabled in critical sections the `wait_for` also does not work.

Is it really needed to have all these things inside of a critical section?
Or should we maybe turn the critical section of when doint `wait_for` and turn it on after?

I also have some other issues caused by this: I now create a thread to manage the connections of all the CoAP channels.
But that thread is created in the constructor of the Coap channel. But that is also within the critical section, so the thread is not created until the critical section is closed.

But the problem is the critical section cannot be closed because it first waits until all channels are connected.

But that is not possible because the coap thread for handling the connection has not started yet. So we are going in circles :D

erlingrj:
> @erlingrj So I found out why the `wait_for` doesn't work here. The `wait_for` is used within a critical section inside `lf_start` and since interrupts are disabled in critical sections the `wait_for` also does not work.
> 

That's a great find! Important detail that `PlatformRiot_wait_for` does not work when called from a critical section. This might cause problems in the future also. Either we have to decide that `wait_for` cannot be called from a critical_section and possibly test this and throw an error. Or we have to use a different underlying RIOT API for wait_for. I think we should go with the latter, make sure that `wait_for` also can be called from a critical section on RIOT.

> Is it really needed to have all these things inside of a critical section? Or should we maybe turn the critical section of when doint `wait_for` and turn it on after?
> 

I think we can safely reduce the critical_section here. I am trying to remember why I added it. It must've been to protect against the NetworkChannel thread doing anything inside the runtime (which would be through the callback function passed to it). This callback `FederatedConnectionBundle_msg_received_cb` in `federated.c`. 

I think we can safely remove the critical section but we have to be careful with the order in which things are done:

1. Construct federate and reactors, this will spawn any network threads which will be busy-spinning waiting for `open_connection()`.
2. Setup `net_bundles` pointer on `Environment` to point to `_bundles` on federated reactor
3. Call `Environment_assemble()`, this will release the network threads.
4. Call `Environment_start()` to wait for StartTagSignal and start the scheduler.

It is important that (2) comes before (3). 

> I also have some other issues caused by this: I now create a thread to manage the connections of all the CoAP channels. But that thread is created in the constructor of the Coap channel. But that is also within the critical section, so the thread is not created until the critical section is closed.
> 
> But the problem is the critical section cannot be closed because it first waits until all channels are connected.
> 
> But that is not possible because the coap thread for handling the connection has not started yet. So we are going in circles :D

That is a good point! I think we can remove the critical section if we are careful in constructing the `lf_start()` function

examples/riot/coap_federated/receiver/Makefile
erlingrj, line 1: Thanks for the example, but can you add a short README for how to run it, including setting up the TAP interface. I have done this, but I suspect I need to pass some argument to the Makefile because both receiver and sender tries to open tap0

src/platform/riot/riot.c
```diff
@@ -64,12 +64,12 @@ lf_ret_t PlatformRiot_wait_for(Platform *self, interval_t duration) {
 
 void PlatformRiot_leave_critical_section(Platform *self) {
   PlatformRiot *p = (PlatformRiot *)self;
-  p->irq_mask = irq_disable();
+  irq_restore(p->irq_mask);
```
erlingrj, line 67: That's a great catch.  Leaving the critical section means enabling IRQs. A little surprised this did not cause any issues earlier...

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -26,21 +26,31 @@ static void _CoapUdpIpChannel_update_state(CoapUdpIpChannel *self, NetworkChanne
...
+  mutex_unlock(&self->state_mutex);
 
   // Inform runtime about new state if it changed from or to NETWORK_CHANNEL_STATE_CONNECTED
   if ((old_state == NETWORK_CHANNEL_STATE_CONNECTED && new_state != NETWORK_CHANNEL_STATE_CONNECTED) ||
       (old_state != NETWORK_CHANNEL_STATE_CONNECTED && new_state == NETWORK_CHANNEL_STATE_CONNECTED)) {
     _env->platform->new_async_event(_env->platform);
   }
+
+  // Let connection thread evaluate new state of this channel
+  msg_t msg = {
```
erlingrj, line 40: So this writes a "msg" to the network thread to notify that the state has been updated. This implies that both runtime and network thread might update the state, I think we discussed earlier to just have the network thread update the state, but this was perhaps only possible for the TcpIp channel?

erlingrj (approved):
This looks great, if you could just provide some info for how to run the example we can merge this! :+1:

src/platform/riot/riot.c
LasseRosenow: Thanks for the feedback, I have added a readme to the examples/riot folder and added entries for how to run the various examples.

We currently also have examples in the main readme.md file, but I didn't  want to make it too overly crowded there. Maybe we should link from the root readme.md file to this new file for RIOT examples and also add examples in folders of the other platforms?

LasseRosenow: Ah I forgot about one thing. The IP Addresses are not always the same at least per host. So I will add that also to the readme quickly

LasseRosenow:
Okay I have added a readme to the examples folder that explains how to build the various examples including the federated coap example.

LasseRosenow:
I am currently also in parallel working on a test that we can run in the CI that uses the tap interface and sends some messages between two devices, but I think we should merge this first. I still have some issues with the  test that I need to figure out

examples/riot/README.md
```diff
@@ -0,0 +1,78 @@
...
+Enter the directory of the `sender` application:
+
+```shell
+cd coap_federated/sender
+```
+
+Get the IP address of the `receiver` by specifying the `PORT=tap1` and `ONLY_PRINT_IP=1` environment variables:
+
+```shell
+make ONLY_PRINT_IP=1 BOARD=native PORT=tap1 all term
```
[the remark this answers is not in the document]
erlingrj: It should be `ONLY_GET_IP`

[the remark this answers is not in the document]
erlingrj: Still confused by the output:
```sh
make ONLY_GET_IP=1 BOARD=native PORT=tap1 term
/home/erling/tools/RIOT/dist/tools/pyterm/pyterm -ps /home/erling/dev/reactor-uc/examples/riot/coap_federated/sender/bin/native/lf-coap-federated-sender.elf --process-args tap1 
Twisted not available, please install it if you want to use pyterm's JSON capabilities
Welcome to pyterm!
Type '/exit' to exit.
2025-01-08 17:00:25,982 # RIOT native interrupts/signals initialized.
2025-01-08 17:00:25,983 # RIOT native board initialized.
2025-01-08 17:00:25,983 # RIOT native hardware initialization complete.
2025-01-08 17:00:25,983 # 
2025-01-08 17:00:25,983 # main(): This is RIOT! (Version: 2024.10)
2025-01-08 17:00:25,983 # [INFO] [NET] IPv6 address: fe80::b020:26ff:fef2:2497
2025-01-08 17:00:25,983 # [INFO] [NET] IPv6 address: f4ff:708:f4ff:708:60bd:608:f3:41dc
2025-01-08 17:00:25,983 # [DEBUG] [NET] CoapUdpIpChannel: Start connection thread
2025-01-08 17:00:25,983 # [INFO] [NET] CoapUdpIpChannel: Register receive callback
2025-01-08 17:00:25,983 # [DEBUG] [ENV] Calculating levels for Reactor MainSender
2025-01-08 17:00:25,983 # [DEBUG] [ENV] Calculating levels for Reactor Sender
2025-01-08 17:00:25,983 # [DEBUG] [ENV] Reaction 0 has level 0
2025-01-08 17:00:25,983 # [DEBUG] [NET] CoapUdpIpChannel: Open connection
2025-01-08 17:00:25,983 # [DEBUG] [NET] CoapUdpIpChannel: Update state: UNINITIALIZED => OPEN
2025-01-08 17:00:25,984 # 
2025-01-08 17:00:25,984 # [DEBUG] [NET] CoapUdpIpChannel: Sending 14 bytes
2025-01-08 17:00:25,984 # [DEBUG] [NET] CoapUdpIpChannel: CoAP Message sent
2025-01-08 17:00:25,984 # [DEBUG] [NET] CoapUdpIpChannel: Update state: OPEN => CONNECTION_IN_PROGRESS
2025-01-08 17:00:25,984 # 
2025-01-08 17:00:27,386 # [DEBUG] [NET] CoapUdpIpChannel: Client open connection callback
2025-01-08 17:00:27,387 # [ERROR] [NET] CoapUdpIpChannel: TIMEOUT => Try to connect again
2025-01-08 17:00:27,387 # [DEBUG] [NET] CoapUdpIpChannel: Update state: CONNECTION_IN_PROGRESS => CONNECTION_FAILED
2025-01-08 17:00:27,387 # 
2025-01-08 17:00:27,387 # [DEBUG] [NET] CoapUdpIpChannel: Sending 14 bytes
2025-01-08 17:00:27,387 # [DEBUG] [NET] CoapUdpIpChannel: CoAP Message sent
2025-01-08 17:00:27,387 # [DEBUG] [NET] CoapUdpIpChannel: Update state: CONNECTION_FAILED => CONNECTION_IN_PROGRESS
2025-01-08 17:00:27,387 # 
2025-01-08 17:00:29,773 # [DEBUG] [NET] CoapUdpIpChannel: Client open connection callback
2025-01-08 17:00:29,773 # [ERROR] [NET] CoapUdpIpChannel: TIMEOUT => Try to connect again
2025-01-08 17:00:29,773 # [DEBUG] [NET] CoapUdpIpChannel: Update state: CONNECTION_IN_PROGRESS => CONNECTION_FAILED

...
```

examples/riot/coap_federated/sender/Makefile
```diff
@@ -0,0 +1,43 @@
...
+# Configure CoAP retransmission timeout
+CFLAGS += -DCONFIG_GCOAP_NO_RETRANS_BACKOFF=1
+CFLAGS += -DCONFIG_COAP_ACK_TIMEOUT_MS=400
+CFLAGS += -DCONFIG_COAP_MAX_RETRANSMIT=4
+
+# Set the remote receiver address
+ifndef REMOTE_ADDRESS
+ifndef ONLY_GET_IP
+  $(error REMOTE_ADDRESS is not defined. Please define it!)
+else
```
erlingrj, line 37: Is this correct Make syntax?

examples/riot/coap_federated/receiver/Makefile
```diff
@@ -0,0 +1,43 @@
...
+# Configure CoAP retransmission timeout
+CFLAGS += -DCONFIG_GCOAP_NO_RETRANS_BACKOFF=1
+CFLAGS += -DCONFIG_COAP_ACK_TIMEOUT_MS=400
+CFLAGS += -DCONFIG_COAP_MAX_RETRANSMIT=4
+
+# Set the remote sender address
+ifndef REMOTE_ADDRESS
+ifndef ONLY_GET_IP
+  $(error REMOTE_ADDRESS is not defined. Please define it!)
+else
```
erlingrj, line 37: Is this correct Make?

erlingrj (changes requested):
Thanks for providing the README, I was still not able to follow it, but it is probably just some minor confusion.

examples/riot/coap_federated/sender/Makefile
LasseRosenow: I have rewritten this. I think it is allowed, but I made it unnecessarily complicated.

examples/riot/coap_federated/receiver/Makefile
LasseRosenow: I have rewritten this. I think it is allowed, but I made it unnecessarily complicated.

src/platform/riot/coap_udp_ip_channel.c
LasseRosenow: Yes the issue here is that the TcpIpChannel is `polling`. We have a main loop and wait for incoming messages and then depending on what comes in we update the state.

But the RIOT `CoAPChannel` is async. Messages come in from the internal `gcoap thread`.
So we have a lot of callbacks such as `_client_open_connection_callback` etc.
And there is also one place where the runtime changes state which is in `open_connection`. This also exists in the current implementation of the TcpIpChannel from what I can see. So if we want to really set it only from one thread maybe we need  to fix this also.

A solution here could be to maybe have an extra state variable that is only set by the runtime. (connection_open: bool) or something like that. This one would only be read by the coap / tcpip thread and then they would set their internal state accordingly?

I opened a new Issue (#183). I think this PR is already quite large and since this anyways not only affects the Coap channel alone, maybe we can fix this in a follow up PR?

erlingrj: Thanks for the clarification. Yes `open_connection` is there, in the Federated branch I use a mutex now to protect the shared state. I think that is necessary

erlingrj:
OK I have it running now! It works please check my updates in the README. 

Hopefully we can find a less manual and error prone way of spinning them up. Easy to mix up how to set `PORT=tap0` or `tap1` which changes depending on `sender`, `receiver` and whether you launch it or you want to find the IP address of the other.

erlingrj (approved):

LasseRosenow:
> Hopefully we can find a less manual and error prone way of spinning them up. Easy to mix up how to set PORT=tap0 or tap1 which changes depending on sender, receiver and whether you launch it or you want to find the IP address of the other.

I 100% agree! :)

LasseRosenow:
Your changes look good, thank you!

PR #143, opened 2024-11-29
Remove duplicated NetworkChannel return codes / state

LasseRosenow:
Ready to review


LasseRosenow:
Okay I think this should be ready for review now :) I am not entirely sure if the runtime internal stuff is implemented correctly.
I will also work on a follow-up PR now that simplifies the NetworkChannel interface as discussed and moves more into the channels internally.

src/federated.c
```diff
@@ -19,21 +19,22 @@ void FederatedConnectionBundle_connect_to_peers(FederatedConnectionBundle **bund
...
       FederatedConnectionBundle *bundle = bundles[i];
       NetworkChannel *chan = bundle->net_channel;
-      NetworkChannelState state = chan->get_connection_state(chan);
-      if (state != NETWORK_CHANNEL_STATE_CONNECTED) {
-        ret = chan->try_connect(chan);
-        switch (ret) {
-        case LF_OK:
-          bundle->network_channel_state_changed(bundle);
+      ret = chan->try_connect(chan);
+      if (ret == LF_OK) {
```
tanneberger, line 29: Shouldn't also the channel state be updated if try_connect failes?

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -7,17 +7,15 @@
 #include "net/sock/util.h"
 #include <arpa/inet.h>
 
-static bool _is_coap_initialized = false;
+static bool _is_globals_initialized = false;
```
tanneberger, line 10: we need those globals for some kind of handler function?

src/platform/posix/tcp_ip_channel.c
```diff
@@ -198,7 +210,7 @@ static lf_ret_t TcpIpChannel_check_socket_error(int fd) {
   }
 }
 
-static lf_ret_t TcpIpChannel_try_connect_client(NetworkChannel *untyped_self) {
+static lf_ret_t _try_connect_client(NetworkChannel *untyped_self) {
```
tanneberger, line 213: the underscore is indicating that this is a "private" member function?

LasseRosenow: Yes as far as I understood it the TcpIpChannel Prefix is to mark the functions that implement the interface, so all other internal code structuring functions are easier to differentiate with for example included functions by using underscore to mark them as private

src/platform/riot/coap_udp_ip_channel.c
LasseRosenow: Yes I need `Environment` to get the channel bundle list in most callbacks to find the channel based on its socket. (See get_channel_by_remote or similar?)

src/federated.c
LasseRosenow: Actually I am not sure because the channel itself should have already notified the bundle, I would need to look deeper into the runtime code tomorrow to see if that is covered

src/platform/posix/tcp_ip_channel.c
```diff
@@ -365,15 +375,20 @@ lf_ret_t TcpIpChannel_receive(NetworkChannel *untyped_self, FederateMessage *ret
...
         break;
       }
       continue;
     } else if (bytes_read == 0) {
       // This means the connection was closed.
-      self->state = NETWORK_CHANNEL_STATE_LOST_CONNECTION;
-      return LF_CONNECTION_CLOSED;
+      LF_INFO(NET, "Connection closed. Setting last known tag to FOREVER for all input ports");
+      if (self->federated_connection) {
+        self->federated_connection->network_channel_state_changed(self->federated_connection);
```
erlingrj, line 387: If I remember correctly, we talked about removing this call altogether and rely on the runtime checking the network channel state before waiting for input ports to become known at a tag.

src/federated.c
```diff
@@ -108,14 +109,8 @@ void FederatedOutputConnection_cleanup(Trigger *trigger) {
 
     LF_DEBUG(FED, "FedOutConn %p sending tagged message with tag=%" PRId64 ":%" PRIu32, trigger, tagged_msg->tag.time,
              tagged_msg->tag.microstep);
-    ret = channel->send_blocking(channel, &msg);
-    switch (ret) {
-    case LF_OK:
-      break;
-    case LF_CONNECTION_CLOSED:
+    if (channel->send_blocking(channel, &msg) != LF_OK) {
       self->bundle->network_channel_state_changed(self->bundle);
```
erlingrj, line 113: If we remove this "callback"-way of handling network-channel state updates (see my other comment), then this call should probably also be dropped. Instead we check the network channel state each time before going to sleep waiting for some input port to become known at a tag

src/platform/posix/tcp_ip_channel.c
erlingrj: I am fine with this convention. But I would also be fine with omitting the `_`, since the functions are static they will not be visible outside this file anyways.

erlingrj (commented):
This looks good, I believe this is the first of a two-stage change and I think my questions are both things that will be done in the next stage?

src/platform/posix/tcp_ip_channel.c
LasseRosenow: For me the main advantage of it is to help differentiate between our own static functions and for example included library functions. I can directly see that for example `send_coap_message` is a private function and not a RIOT coap API function.

But I also don't have a strong opinion on it

LasseRosenow: I mean within that c file of course

src/federated.c
LasseRosenow: Regarding all the callback comments I think that change should go into the second PR. I tried to keep the scope of this one only within the deduplication of the return code and state :)

src/platform/posix/tcp_ip_channel.c
tanneberger: how about:

`_<Class-Name>_<Member-Name>` for private member functions
`<Class-Name>_<Member-Name>` for public member functions

... would be a bit more consistent. Then the leading `_` would clearly indicate visibility.

LasseRosenow: What I am not sure is would you consider function such as `TcpIpChannel_send_blocking` as private or as public?

From the code perspective they are `private`, but then exposed through the super type `NetworkChannel`.
Still different though compared to for example to `TcpIpChannel_ctor` which is indeed `public`.

So do we say:
- `TcpIpChannel_ctor` (public)
- `TcpIpChannel_send_blocking` (private, but exposed through super type)
- `_TcpIpChannel_some_internal_helper_function' (private)

Or:
- `TcpIpChannel_ctor` (public)
- `_TcpIpChannel_send_blocking` (private, but exposed through super type)
- `_TcpIpChannel_some_internal_helper_function' (private)

I personally think prefixing all the internal functions with `_TcpIpChannel_` makes the code somewhat harder to read, but I think it is still acceptable. But we need to make sure we are on the same page on how it is exactly implemented.

src/federated.c
```diff
@@ -280,21 +278,6 @@ void FederatedConnectionBundle_msg_received_cb(FederatedConnectionBundle *self,
...
 }
 
-void FederatedConnectionBundle_network_channel_state_changed(FederatedConnectionBundle *self) {
-  NetworkChannelState state = self->net_channel->get_connection_state(self->net_channel);
-  switch (state) {
-  case NETWORK_CHANNEL_STATE_CONNECTED:
-  case NETWORK_CHANNEL_STATE_LOST_CONNECTION:
-    for (size_t i = 0; i < self->inputs_size; i++) {
-      FederatedInputConnection *input = self->inputs[i];
-      input->last_known_tag = FOREVER_TAG;
```
LasseRosenow, line 295: I think I still don't understand the runtime fully enough, so I am not sure where to set this tag now instead

LasseRosenow:
Okay so I removed the state_changed callback, but there is one unclear thing remaining (see my comment)

src/platform/posix/tcp_ip_channel.c
erlingrj: I would say:

`TcpIpChannel_` means that it is a public facing API through function pointers on a struct, or through the header. `_TcpIpChannel_` means private helper functions

erlingrj: Which is in accordance with the first

erlingrj: If you agree, then could you update the naming so its in accordance with this, and then we can merge this?

LasseRosenow: Yeah no problem

LasseRosenow:
Okay I implemented the naming convention and rebased with main

src/platform/posix/tcp_ip_channel.c
LasseRosenow: Done ;)

src/federated.c
erlingrj: We could set it in Scheduler_acquire_tag, there we can check the state of the input port before we go to sleep waiting for more messages on it. We can add that in the next PR

erlingrj (approved):
Great stuff Lasse!

PR #147, opened 2024-12-02
Simplify NetworkChannel interface

LasseRosenow:
WIP: I only migrated the runtime and CoapUdpIpChannel for now.

Depends on: #143


src/platform/posix/tcp_ip_channel.c
```diff
@@ -21,29 +21,25 @@
...
-  // Set socket to blocking
-  int opts = fcntl(fd, F_GETFL);
-  if (opts < 0) {
-    throw("Could not get socket options");
-  }
-  if (blocking) {
-    opts = (opts & (~O_NONBLOCK));
-  } else {
-    opts = (opts | O_NONBLOCK);
-  }
```
LasseRosenow, line 39: This is not needed anymore since we can just always block inside of the TcpIpChannel now. It also makes handling the state more easy. The runtime is not blocked by it anymore.

LasseRosenow:
This branch diverged too much from the one it depended on. Rebasing proofed difficult so I will reopen a new branch with the changes

PR #150, opened 2024-12-04
Simplify NetworkChannel interface

LasseRosenow:
This pull request removes the `try_connect` function from the `NetworkChannel`.
The runtime will now only tell the `NetworkChannel` implementation such as `CoapUdpIpChannel` to open or close a connection. The channel will then internally take care of connection establishment.

The runtime can get the `NetworkChannel` state using `get_conneciton_state`.


src/platform/posix/tcp_ip_channel.c
```diff
@@ -25,122 +25,106 @@ static bool _is_globals_initialized = false;
...
-  _TcpIpChannel_update_state(self, NETWORK_CHANNEL_STATE_OPEN);
-
   return LF_OK;
 }
 
-static void _TcpIpChannel_spawn_receive_thread(TcpIpChannel *self) {
+static void _TcpIpChannel_spawn_worker_thread(TcpIpChannel *self) {
   int res;
-  LF_INFO(NET, "TCP/IP spawning callback thread");
+  LF_INFO(NET, "TcpIpChannel: Spawning callback thread");
```
erlingrj, line 76:
```suggestion
  LF_INFO(NET, "TcpIpChannel: Spawning worker thread");
```

src/platform/posix/tcp_ip_channel.c
```diff
@@ -405,35 +381,70 @@ static lf_ret_t _TcpIpChannel_receive(NetworkChannel *untyped_self, FederateMess
...
+    LF_ERR(NET, "TcpIpChannel: Error closing server socket %d", errno);
   }
 }
 
-static void *_TcpIpChannel_receive_thread(void *untyped_self) {
-  LF_INFO(NET, "Starting TCP/IP receive thread");
+/**
+ * @brief Main loop of the TcpIpChannel.
+ */
+static void *_TcpIpChannel_worker_thread(void *untyped_self) {
```
erlingrj, line 401: I like this new organization of the code. Great!

src/platform/posix/tcp_ip_channel.c
```diff
@@ -308,12 +284,12 @@ static lf_ret_t TcpIpChannel_send_blocking(NetworkChannel *untyped_self, const F
...
     ssize_t bytes_send = send(socket, self->write_buffer + bytes_written, message_size - bytes_written, 0);
-    LF_DEBUG(NET, "%d bytes sent", bytes_send);
+    LF_DEBUG(NET, "TcpIpChannel: %d bytes sent", bytes_send);
 
     if (bytes_send < 0) {
-      LF_ERR(NET, "write failed errno=%d", errno);
+      LF_ERR(NET, "TcpIpChannel: Write failed errno=%d", errno);
       switch (errno) {
       case ETIMEDOUT:
       case ENOTCONN:
```
erlingrj, line 295: Is it necessary to call `_TcpIpChannel_update_state` here? We might get some very subtle races here when the worker thread also detects that the socket is closed he will initiate a reconnect right? Then we don't want to overwrite the new state if we fail to send some data as well?

If we are guaranteed that the worker thread will not block forever on `receive`, and that it will detect anything that is detected here, then I propose that we only read the channel state from the runtime side, and only write to it from the worker thread. What do you think?

erlingrj (commented):
This is looking very good, I like the new organization a lot. There is an important question about possible race conditions (current and future) wrt to updating the channel state.

LasseRosenow:
@erlingrj I am thinking about one thing: It seems to me that the runtime really only cares about one thing now: Is the channel connected or not. So maybe it is reasonable to replace that `get_connection_state` function with a simple boolean function: `is_connected`?

src/platform/posix/tcp_ip_channel.c
LasseRosenow: Yes there might be a race if the worker thread already started a reconnection. We could guard this though.
I think it is better to update the state straight away if we see something is wrong.

A very robust way could be to just mutex the whole worker thread `while loop` and then also mutex the whole send_blocking function? And of course also the `get_state` function. This way both should not be able to interfer with each other.
We should though then change back to making the tcp `receive` non-blocking.

LasseRosenow: I have now implemented a possible solution in the [Fix race conditions](https://github.com/lf-lang/reactor-uc/pull/150/commits/373782dec4f7a724053d82a2815f691ded470876) commit

LasseRosenow:
I have now also checked that the POSIX federated example works :) So I mark this PR as ready 🥳

src/platform/posix/tcp_ip_channel.c
```diff
@@ -27,13 +27,33 @@ static Environment *_env;
 // Forward declarations
 static void *_TcpIpChannel_worker_thread(void *untyped_self);
 
+static char *_TcpIpChannel_state_to_string(NetworkChannelState state) {
```
erlingrj: Nice!

src/platform/posix/tcp_ip_channel.c
```diff
@@ -411,41 +457,56 @@ static void *_TcpIpChannel_worker_thread(void *untyped_self) {
...
-    } break;
-
-    case NETWORK_CHANNEL_STATE_CONNECTED: {
-      ret = _TcpIpChannel_receive(untyped_self, &self->output);
-      if (ret == LF_OK) {
-        validate(self->receive_callback);
-        self->receive_callback(self->federated_connection, &self->output);
-      } else {
-        LF_ERR(NET, "TcpIpChannel: Error receiving message %d", ret);
+    pthread_mutex_lock(&self->state_mutex);
```
erlingrj, line 439: This means that we cannot send concurrently with receiving? It seems a bit too constraining?

src/platform/posix/tcp_ip_channel.c
```diff
@@ -411,41 +457,56 @@ static void *_TcpIpChannel_worker_thread(void *untyped_self) {
...
+        break;
       }
-    } break;
-
-    case NETWORK_CHANNEL_STATE_UNINITIALIZED:
-    case NETWORK_CHANNEL_STATE_CLOSED:
-      break;
     }
+    pthread_mutex_unlock(&self->state_mutex);
+    _env->platform->wait_for(_env->platform, TCP_IP_CHANNEL_WORKER_THREAD_MAIN_LOOP_SLEEP);
```
erlingrj, line 488: It is a little unfortunate that we have to set a sleep here. The duration of the sleep adds to the worst case latency between two federates. Also it means that we are always polling the the receive socket and that we never really can go to low-power sleep

src/platform/posix/tcp_ip_channel.c
```diff
@@ -261,56 +296,65 @@ static lf_ret_t TcpIpChannel_open_connection(NetworkChannel *untyped_self) {
...
-  }
-
-  // serializing protobuf into buffer
-  int message_size = serialize_to_protobuf(message, self->write_buffer, TCP_IP_CHANNEL_BUFFERSIZE);
-
-  if (message_size < 0) {
-    LF_ERR(NET, "TcpIpChannel: Could not encode protobuf");
-    return LF_ERR;
-  }
+  pthread_mutex_lock(&self->state_mutex);
```
erlingrj, line 280: If the worker thread is constantly polling the socket, then I think he will detect the closed socket very quickly anyways and we would need to handle this from the `send_blocking` side and we wouldn't need to lock any mutex here.

However, I am unsure whether we want to have the worker thread busy-poll the receive socket. Is it possible for the worker thread to both block on the receiving socket AND a condition variable? And then we could signal that condition variable from `send_blocking` if we fail to send. That would wake up the worker thread, and then let it handle the updating of state etc?

erlingrj (commented):
Thanks for addressing my comments Lasse and continuing working on the PR. I am still hoping we can find a more optimal solution for TcpIpChannel. I think you have solved the race conditions, but I think we sacrifice too much. Ideally we can avoid both busy-polling the receive socket, and also allow concurrently receiving and sending data.

We could for instance do a `select` on the receive socket with a timeout of 1s. And if we time-out, then we can grab a mutex. And check if a flag was set from `send_blocking` signalling a problem we transmitting data. And then possibly instigate a reconnection?

LasseRosenow:
Thanks for the feedback! All good points I will try to find a better solution tomorrow

erlingrj:
> @erlingrj I am thinking about one thing: It seems to me that the runtime really only cares about one thing now: Is the channel connected or not. So maybe it is reasonable to replace that `get_connection_state` function with a simple boolean function: `is_connected`?
> 
> Or do you think maybe there could be reasons why we need more information in the future?

Let's do this, I think it is better to keep the API as minimal as possible and then potentially expand later

LasseRosenow:
So I hope I have addressed the parallel send and receive issue now.
I am now using a pipe between `send_blocking` and the `worker thread`.
The `worker thread` uses `select` to listen for the `channel socket` and the `pipe`.
I also fixed some issues with `reset_socket` causing the other federate to close the channel which is something that should only happen from within the runtime in my opinion, because it makes it impossible for the tcp ip channel to recover

LasseRosenow:
I have now changed it to use an eventfd_t instead of a pipe. This way it should be compatible to zephyr OS also

src/platform/posix/tcp_ip_channel.c
```diff
@@ -21,126 +22,115 @@
 
 #define MIN(a, b) ((a) < (b) ? (a) : (b))
 
+#define TCP_IP_CHANNEL_ERR(fmt, ...)                                                                                   \
```
erlingrj, line 25: Nice

src/platform/posix/tcp_ip_channel.c
```diff
@@ -21,126 +22,115 @@
...
-    LF_ERR(NET, "Error setting socket options %d", errno);
+    TCP_IP_CHANNEL_ERR("Error setting socket options: SO_REUSEADDR errno=%d", errno);
     return LF_ERR;
   }
 
-  _TcpIpChannel_socket_set_blocking(self->fd, false);
-  _TcpIpChannel_update_state(self, NETWORK_CHANNEL_STATE_OPEN);
+  self->send_failed_event_fds = eventfd(0, EFD_NONBLOCK);
+  if (self->send_failed_event_fds == -1) {
+    TCP_IP_CHANNEL_ERR("Failed to initialize event file descriptor");
```
erlingrj, line 86:
```suggestion
    TCP_IP_CHANNEL_ERR("Failed to initialize event file descriptor");
    return LF_ERR;
```

src/platform/posix/tcp_ip_channel.c
```diff
@@ -405,35 +337,98 @@ static lf_ret_t _TcpIpChannel_receive(NetworkChannel *untyped_self, FederateMess
...
+      // Set up the file descriptor set
+      FD_ZERO(&readfds);
+      FD_SET(socket, &readfds);
+      FD_SET(self->send_failed_event_fds, &readfds);
+
+      // Determine the maximum file descriptor for select
+      max_fd = (socket > self->send_failed_event_fds) ? socket : self->send_failed_event_fds;
+
+      // Wait for data or cancel if send_failed externally
+      select(max_fd + 1, &readfds, NULL, NULL, NULL);
```
erlingrj, line 409: Should we check the return value of `select`?

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -223,89 +254,68 @@ static lf_ret_t CoapUdpIpChannel_try_connect(NetworkChannel *untyped_self) {
...
 
   // Send message
-  if (_CoapUdpIpChannel_send_coap_message_with_payload(self, &self->remote, "/message",
-                                                       _CoapUdpIpChannel_client_send_blocking_callback, message)) {
-    return LF_OK;
+  self->send_ack_received = false;
+  if (_CoapUdpIpChannel_send_coap_message_with_payload(self, &self->remote, "/message", _client_send_blocking_callback,
+                                                       message)) {
+    // Wait until the response handler confirms the ack or times out
+    while (!self->send_ack_received) {
```
erlingrj, line 309: I guess this is the only way for us to guarantee in-order reliable delivery, right?

erlingrj (approved):
This looks very good to me. Just minor nitpicks. OK to merge after addressing them!

src/platform/riot/coap_udp_ip_channel.c
LasseRosenow: Actually this was implemented to achieve a `blocking` send. Just like in TCP we want to send blocking so we wait for the message to be `acked`.

But you are correct this somewhat also gives us in-order guarantee, but I am not sure if it covers all cases. We can think about this more later I think.

erlingrj: Yes, this is a sane starting point and we can think about edge cases and intermittent connections etc later

LasseRosenow:
The review should be addressed now :)

PR #155, opened 2024-12-05
reactor-c compatibility header

tanneberger:


include/reactor-uc/reactor-c.h
```diff
@@ -0,0 +1,36 @@
+#ifndef REACTOR_C_H
+#define REACTOR_C_H
+
+#define lf_print(str, ...) printf(str, __VA_ARGS__)
+#define PRINTF_TIME "%" PRId64
+#define PRINTF_TAG "(" PRINTF_TIME ", " PRINTF_MICROSTEP ")"
```
erlingrj, line 6: Dont we need to define PRINTF_MICROSTEP also? Should by PRId32 I think

tanneberger: you are right - fixed

erlingrj (changes requested):
LGTM

PR #159, opened 2024-12-08
Have LFC generate a main function that can optionally be included

erlingrj:
Also, build Zephyr and RIOT LF example in CI

### DIFF
--- a/.github/actions/riot/action.yml
+++ b/.github/actions/riot/action.yml
@@ -13,7 +13,7 @@ runs:
     - name: Install RIOT
       if: steps.cache-riot.outputs.cache-hit != 'true'
       run: |
-        git clone -b https://github.com/RIOT-OS/RIOT.git ${{env.RIOTBASE}}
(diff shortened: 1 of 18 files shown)


erlingrj:
@LasseRosenow, can you see why the RIOT build systems fails to include `lf_main.c` even though I explicitly list it as a source in `riot.mk`?

LasseRosenow:
> @LasseRosenow, can you see why the RIOT build systems fails to include `lf_main.c` even though I explicitly list it as a source in `riot.mk`?

I will look at it when I have time later this evening

LasseRosenow:
There must be a CI specific issue, I was able to build it on my machine :O

LasseRosenow:
Ooooohhhh actually I remember why. The problem is in the CI we are currently using riot 2024.07.

Adding c files in folders using SRC += is new since a few weeks ago. Sadly it didn't even make it into RIOT 2024.10.

So we should maybe switch the ci to a specific commit in the master branch for now?

LasseRosenow:
Yay it works now 🥳

erlingrj:
Great! Thank you :100:

PR #166, opened 2024-12-12
RIOT: auto execute LFC on build

LasseRosenow:
This adds the lfc command to `riot.mk` and also moves more configuration over from the lingo template.

For more info see: https://github.com/lf-lang/lf-riot-template/pull/1


make/riot/riot.mk
```diff
@@ -1,12 +1,31 @@
...
+# Check if REACTOR_UC_PATH variable exist
+ifndef REACTOR_UC_PATH
+$(error REACTOR_UC_PATH is not defined. Please define it!)
+endif
 
 # Name of your RIOT application
 APPLICATION ?= $(LF_MAIN)
 
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/RIOT
```
erlingrj, line 21: This assumes that you are building your application from our lf-riot-template and have RIOT as a submodule. Ideally we would keep such assumptions in the makefile within `lf-riot-template`. IIRC we discussed moving some of this back to the makefile in `lf-riot-template`, did you request my review before or after the meeting :stuck_out_tongue:

LasseRosenow: Yeah that was before the meeting haha :)

LasseRosenow: Done

LasseRosenow:
@erlingrj Ready for review again :)

examples/riot/buildAll.sh
```diff
@@ -21,8 +21,3 @@ for board in "${BOARDS[@]}"; do
 			popd
 	done
 done
-
-# Build lf example
-pushd hello_lf
-run/build.sh
-popd
```
LasseRosenow, line 28: Integrating the LFC into the make file is really nice! Now we don't need to handle _LF examples different from c-code based examples anymore!

examples/riot/hello_lf/Makefile
```diff
@@ -1,14 +1,20 @@
...
+
+# Enable reactor-uc features
+# CFLAGS += -DNETWORK_CHANNEL_TCP_POSIX
+# CFLAGS += -DNETWORK_CHANNEL_COAP_RIOT
 
-# Path of generated lf c-code
-LF_SRC_GEN_PATH = $(CURDIR)/src-gen/$(LF_MAIN)
+# Execute the LF compiler if build target is "all"
+ifeq ($(MAKECMDGOALS),all)
+  _ :=  $(shell $(REACTOR_UC_PATH)/lfc/bin/lfc-dev src/$(LF_MAIN).lf)
```
erlingrj, line 12: Just for my own understanding, what does this syntax here achieve? Will the command be invoked before any target is run?

make/riot/riot-lfc.mk
```diff
@@ -0,0 +1,26 @@
...
+# Check if required environment variables exist
+ifndef LF_MAIN
+  $(error LF_MAIN is not defined. Please define it!)
+endif
+
+# Name of your RIOT application
+APPLICATION ?= $(LF_MAIN)
+
+# Path of generated lf c-code
+LF_SRC_GEN_PATH ?= $(CURDIR)/src-gen/$(LF_MAIN)
```
erlingrj, line 12: An unfortunate detail about lfc is that where the generated-code is put is a little complicated:
1. `~/src/Hello.lf` -> `~/src-gen/Hello`
2. `~/Hello.lf` -> `~/src-gen/Hello`
3. `~/src/lib/Hello.lf` -> `~/src-gen/lib/Hello

I think we should ignore (2) where it is not in a `src` directory, for our template we say that the LF files must reside under a `src` directory. But for (3) to work, LF_MAIN must be set to `lib/Hello`, which might actually be reasonable

make/riot/riot-lfc.mk
```diff
@@ -0,0 +1,26 @@
...
+# Include generated c files
+SRC += $(patsubst %, $(LF_SRC_GEN_PATH)/%, $(LFC_GEN_SOURCES))
+
+# Include generated main file
+SRC += $(LF_SRC_GEN_PATH)/${LFC_GEN_MAIN}
+
+# Include generated h files
+CFLAGS += -I$(LF_SRC_GEN_PATH)
+
+include $(RIOT_MK_DIR)/riot.mk
```
erlingrj, line 26:
```suggestion
include $(RIOT_MK_DIR)/riot.mk

```

erlingrj (approved):
This looks good. Just a few minor comments. I will let you merge it when you've seen them

examples/riot/hello_lf/Makefile
LasseRosenow: This is a small hack, it copies the output of the shell script over to the variable in this case underscore. This way the evaluation of the makefile is blocked until the script is finished ;)

make/riot/riot-lfc.mk
LasseRosenow: Aaahh that is good to know, I will think about a good solution

examples/riot/hello_lf/Makefile
erlingrj: That's very clever!

make/riot/riot-lfc.mk
LasseRosenow: GitHub mobile doesn't let me approve this there is a strange error message. Feel free to just commit it :)

PR #175, opened 2024-12-13
Code-generate for federated programs

erlingrj:
TODOs:
- [x] Nasty FlexPRET bug? 
- [x] Dont generate federate launch script unless we are targeting native
- [x] Have LFC generate binaries for each federate into `bin` together with a simple shell script to invoke them (just like reactor-c)
- [x] We should likely create a UcFederatedPlatformGenerator which is similar to UcStandaloneGenerator but fixes all of this for federated programs
- [x] Main federate should likely be something like `SimpleFederate_r` for a program called `SimpleFederate.lf` and a federate instance called `r`.
- [x] Consider doing all federated code-gen first, and then invoke CMake (now we do code-gen -> cmake , code-gen -> cmake). 
- [x] The sources variable seems to be global and shared between federates. Investigate
- [x] Handle federated connections
- [x] Disallow any reactions in top-level
- [x] Propose an initial attribute syntax for federated connections


examples/riot/buildAll.sh
```diff
@@ -21,3 +21,8 @@ for board in "${BOARDS[@]}"; do
 			popd
 	done
 done
+
+# Build lf example
```
erlingrj, line 25: @LasseRosenow Is this needed?

include/reactor-uc/environment.h
```diff
@@ -9,6 +9,7 @@
 #include "reactor-uc/scheduler.h"
 
 typedef struct Environment Environment;
+extern Environment *_lf_environment;
```
erlingrj, line 12: I created this global backdoor so that the logging functions could read out the current time and add it to the printouts... we could possibly use this also within the TcpIpChannel and UdpCoapChannel

CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ if (NOT PLATFORM STREQUAL "ZEPHYR")
 endif()
 
 # Add compile definitions for platform and network channel specifics.
-target_compile_definitions(reactor-uc PRIVATE "PLATFORM_${PLATFORM}")
+target_compile_definitions(reactor-uc PUBLIC "PLATFORM_${PLATFORM}")
```
erlingrj, line 81: The PLATFORM compile def must be visible in targets that link to reactor-uc also, because they will include the network_channel.h where it is used

src/environment.c
```diff
@@ -49,7 +49,11 @@ interval_t Environment_get_physical_time(Environment *self) {
   return self->platform->get_physical_time(self->platform);
 }
 interval_t Environment_get_elapsed_physical_time(Environment *self) {
-  return self->platform->get_physical_time(self->platform) - self->scheduler->start_time;
+  if (self->scheduler->start_time == NEVER) {
```
erlingrj, line 52: Support reading out elapsed physical time even before we have started

src/federated.c
```diff
@@ -150,7 +155,7 @@ void FederatedInputConnection_ctor(FederatedInputConnection *self, Reactor *pare
   self->delay = delay;
   self->type = type;
   self->last_known_tag = NEVER_TAG;
-  self->safe_to_assume_absent = FOREVER;
+  self->safe_to_assume_absent = 0; // FIXME: This should be set by the user
```
erlingrj, line 158: Note that I change the default here. By setting this to 0 we move from a dataflow federation to a "poor-mans-ptides" where the after-delays on the federated connections should be greater than the actual network delays

src/federated.c
```diff
@@ -225,21 +230,29 @@ void FederatedConnectionBundle_handle_tagged_msg(FederatedConnectionBundle *self
       ret = sched->schedule_at_locked(sched, &event);
       switch (ret) {
       case LF_AFTER_STOP_TAG:
-        LF_WARN(FED, "Tried scheduling event after stop tag. Dropping\n");
+        LF_WARN(FED, "Tried scheduling event after stop tag. Dropping");
         break;
       case LF_PAST_TAG:
-        LF_WARN(FED, "Tried scheduling event to a past tag. Dropping\n");
+        LF_ERR(FED, "Safe-to-process violation! Tried scheduling event to a past tag. Handling now instead!");
```
erlingrj, line 236: If we receive an event destined for a past tag, then instead of dropping it, we handle it ASAP. In the future we will add support for a handler for this case, just like the deadlines

src/platform/posix/posix.c
```diff
@@ -29,7 +29,8 @@ static struct timespec convert_ns_to_timespec(instant_t time) {
 }
 
 lf_ret_t PlatformPosix_initialize(Platform *_self) {
-  signal(SIGINT, handle_ctrlc);
+  signal(SIGINT, handle_signal);
```
erlingrj, line 32: Handle both ctrl-c and SIGTERM and stop the program

examples/riot/buildAll.sh
LasseRosenow: I think this should be fixed when you rebase with main. This does not exist anymore :)

include/reactor-uc/environment.h
LasseRosenow: So you mean we remove the `Environment pointer` from the constructor and just use the global pointer?

erlingrj: Yes that is possible. It is a little dirty, but, having global access to the environment also makes sense.

examples/riot/buildAll.sh
erlingrj: I did a merge with main and ended up with this, will remove it

erlingrj:
@tanneberger and @LasseRosenow this is a big PR and a big task to review it. But I hope you will take a look and give some feedback. Unfortunately the code-generator is still a little quirky trying to fit this federated generation into the existing non-federated architecture. But I think it is better to get this merged now and improve it as we go

include/reactor-uc/network_channel.h
```diff
@@ -88,4 +95,35 @@ struct NetworkChannel {
...
+#endif
+
+#elif defined(PLATFORM_FLEXPRET)
+#ifdef NETWORK_CHANNEL_TCP_POSIX
+#error "NETWORK_POSIX_TCP not supported on FlexPRET"
+#endif
+
+#else
+#error "Platform not supported"
+#endif
```
LasseRosenow, line 128: I like this. Now I don't need to always change the `include` :)

src/platform/posix/tcp_ip_channel.c
```diff
@@ -70,6 +80,18 @@ static lf_ret_t _TcpIpChannel_reset_socket(TcpIpChannel *self) {
       return LF_ERR;
     }
   }
+  if (self->send_failed_event_fds[0] > 0) {
+    if (close(self->send_failed_event_fds[0]) < 0) {
```
LasseRosenow, line 84: Nice catch!

LasseRosenow (commented):
Very nice! I have looked at the code so far and only have a few small questions :)

I will try to get a federated RIOT Coap example running and then approve :)

examples/riot/hello_lf/run/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/env bash
+
+${REACTOR_UC_PATH}/lfc/bin/lfc-dev src/HelloLF.lf
+make all
```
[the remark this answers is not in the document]
erlingrj: Nice catch, this should not be here

include/reactor-uc/logging.h
```diff
@@ -106,7 +117,7 @@
     }                                                                                                                  \
   } while (0)
 
-#define LF_ERR(module, fmt, ...) LF_LOG(LF_LOG_LEVEL_ERR, module, fmt, ##__VA_ARGS__)
+#define LF_ERR(module, fmt, ...) LF_LOG(LF_LOG_LEVEL_ERROR, module, fmt, ##__VA_ARGS__)
 #define LF_WARN(module, fmt, ...) LF_LOG(LF_LOG_LEVEL_WARN, module, fmt, ##__VA_ARGS__)
```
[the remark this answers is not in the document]
erlingrj: For compatability with the `logging: ` target property defined for reactor-c. I guess you could argue that `WARN` is not a abbreviation, but a word in itself

include/reactor-uc/environment.h
```diff
@@ -9,6 +9,7 @@
 #include "reactor-uc/scheduler.h"
 
 typedef struct Environment Environment;
+extern Environment *_lf_environment; // NOLINT
```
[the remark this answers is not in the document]
erlingrj: OK, the I use this environment in TcpIpChannel and CoapUdpChannel in this PR

test/lf/src/only_build/FederatedCoap.lf
```diff
@@ -0,0 +1,34 @@
...
+
+federated reactor {
+
+  @interface_coap(name="if1", address="127.0.0.1")
+  r1 = new Src(id=42)
+
+  @interface_coap(name="if1", address="127.0.0.1")
+  r2 = new Dst()
+
+  @link(left="if1", right="if1", server_side="right")
```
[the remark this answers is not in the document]
LasseRosenow: Right now it is completely symmetric yes

PR #177, opened 2024-12-14
Setup Doxygen

erlingrj:
This PR sets up Doxygen for auto generating documentation. It also adds a Github Actions workflow for deploying the docs to the associated Github pages. 

I also improved the documentation a few places


erlingrj:
This PR publishes the entire Doxygen, not only user-facing. We can consider just publishing the user-facing stuff also

erlingrj:
Check it out here: https://www.lf-lang.org/reactor-uc/

tanneberger:
@erlingrj if we want to use a Doxygen, let's make sure it looks like from this century: https://jothepro.github.io/doxygen-awesome-css/

PR #178, opened 2024-12-16
Migrate eventfd to socketpair to support macOS

### LINKED ISSUE #176: make test is broken
On my Mac when running `make test` I get:
```
/Users/martin/source/lf-world/reactor-uc/src/platform/posix/tcp_ip_channel.c:15:10: fatal error: 'sys/eventfd.h' file not found
#include <sys/eventfd.h>
```

LasseRosenow:
Fixes #176

### DIFF
--- a/src/platform/posix/tcp_ip_channel.c
+++ b/src/platform/posix/tcp_ip_channel.c
@@ -81,9 +81,8 @@ static lf_ret_t _TcpIpChannel_reset_socket(TcpIpChannel *self) {
     return LF_ERR;
   }
 
-  self->send_failed_event_fds = eventfd(0, EFD_NONBLOCK);
-  if (self->send_failed_event_fds == -1) {
-    TCP_IP_CHANNEL_ERR("Failed to initialize event file descriptor");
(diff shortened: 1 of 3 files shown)


src/platform/posix/tcp_ip_channel.c
```diff
@@ -244,7 +243,7 @@ static lf_ret_t TcpIpChannel_send_blocking(NetworkChannel *untyped_self, const F
           switch (errno) {
           case ETIMEDOUT:
           case ENOTCONN:
-            ssize_t bytes_written = eventfd_write(self->send_failed_event_fds, 1);
+            ssize_t bytes_written = write(self->send_failed_event_fds[1], "X", 1);
```
tanneberger, line 246: does `eventfd_write` really write an "X" ?

LasseRosenow: Actually it doesn't matter what you write you just need to write something and x is just one byte

erlingrj (approved):
Thanks for addressing this so quickly, it looks good to me!

PR #189, opened 2025-01-15
Add RIOT federated CoAP TEST

LasseRosenow:
No real test yet, but at least I finally managed to run the sender and the receiver in parallel from one script file.
I had a strange race condition in RIOT perhaps? The sender and receiver were unable to talk over the tap interface, if they started to fast in parallel.

The solution for now is: `sleep 3` .... But it wörks :)

image


erlingrj:
Looks promising. It would be really great to understand what this race condition in the startup is about. Adding explicit sleeps to enforce some ordering often works great until we have some freak CI run where the timing is completely messed up because it is running in the cloud and maybe the CI task was suspended at the wrong point etc...

LasseRosenow:
Okay this is pretty much done. The only problem that remains is that I am having issues setting up the tap interfaces in the CI. I will look more into it

LasseRosenow:
Holy f** the CI has finally passed :D

LasseRosenow:
Finally I got the tap interfaces working in the CI, this is ready for review!! 🥳

examples/riot/coap_federated/receiver/Makefile
```diff
@@ -29,17 +29,8 @@ CFLAGS += -DCONFIG_GCOAP_NO_RETRANS_BACKOFF=1
...
-else ifdef REMOTE_ADDRESS
-  # REMOTE_ADDRESS is defined => Set CFLAGS for it
-  CFLAGS += -DREMOTE_ADDRESS=\"$(REMOTE_ADDRESS)\"
-else
-  # Neither is defined
-  $(error Either define REMOTE_ADDRESS or set ONLY_PRINT_IP=1 to print the IP-Address of this device.)
-endif
+# Static IPv6 address
+CFLAGS += -DCONFIG_GNRC_IPV6_STATIC_LLADDR='"fe80::cafe:cafe:cafe:2"'
+CFLAGS += -DCONFIG_GNRC_IPV6_STATIC_LLADDR_IS_FIXED=1
```
erlingrj, line 34: So these two lines is the way of specifying which IP address the current program should use?

examples/riot/coap_federated/receiver/main.c
```diff
@@ -1,9 +1,7 @@
 #include "reactor-uc/platform/riot/coap_udp_ip_channel.h"
 #include "reactor-uc/reactor-uc.h"
 
-#ifndef REMOTE_ADDRESS
-#define REMOTE_ADDRESS "fe80::44e5:1bff:fee4:dac8"
-#endif
+#define REMOTE_ADDRESS "fe80::cafe:cafe:cafe:1"
```
erlingrj, line 4: Since we specify our own IP address in the Makefile, perhaps we should also specify the other IP address there as well, and not in the source file?

test/platform/riot/coap_channel_federated_test/sender/Makefile
```diff
@@ -0,0 +1,41 @@
...
+# Comment this out to disable code in RIOT that does safety checking
+# which is not needed in a production environment but helps in the
+# development process:
+DEVELHELP ?= 1
+
+# Change this to 0 show compiler invocation lines by default:
+QUIET ?= 1
+
+# Enable Unity
+USEMODULE += Unity
```
erlingrj, line 19: The unit test framework? Is it used for anything?

test/platform/riot/coap_channel_federated_test/run.sh
erlingrj, line 1: I am a little worried on the dependence on time here, sometimes CI runs have very strange timing. But lets see!

erlingrj (approved):
THis looks great! Added a commit fixing the example program. And a few comments

examples/riot/coap_federated/receiver/Makefile
LasseRosenow: This is one way of specifying a specific static IPv6 address in RIOT, but if you want to specify multiple addresses you need to do it using `c-code`

examples/riot/coap_federated/receiver/main.c
LasseRosenow: Fixed

test/platform/riot/coap_channel_federated_test/sender/Makefile
LasseRosenow: No this was a leftover, I fixed it :)

test/platform/riot/coap_channel_federated_test/run.sh
LasseRosenow: Yes I am as well, but I also didn't see a better way, so lets hope it works well :)

LasseRosenow:
Okay I think all the feedback is addressed, thank you!

PR #191, opened 2025-01-16
UARTChannel 

tanneberger:


include/reactor-uc/network_channel.h
```diff
@@ -31,14 +31,27 @@ typedef enum {
 typedef enum {
   NETWORK_CHANNEL_TYPE_TCP_IP,
   NETWORK_CHANNEL_TYPE_COAP_UDP_IP,
+  NETWORK_CHANNEL_TYPE_I2C,
```
erlingrj, line 34: One thing that will make I2C challenging is that we currently require channels to be two-way. This is needed even if there is only a connection going one way. The reason is that we distribute the start tags by forwarding it on all channels. If we only could forward it on outgoing connections we would need to carefully pick the federate selecting the start tag.

include/reactor-uc/network_channel.h
```diff
@@ -31,14 +31,27 @@ typedef enum {
...
+  NETWORK_CHANNEL_TYPE_UART
 } NetworkChannelType;
 
+typedef enum {
+  NETWORK_CHANNEL_CATEGORY_ASYNC,
+  NETWORK_CHANNEL_CATEGORY_SYNC,
+} NetworkChannelCategory;
```
erlingrj, line 41: I think `Category` is a little too vague. What about `Mode`, e.g. `NetworkChannelMode`. But since we only have two modes, we could also just use a boolean flag, like: `boolean polled`

src/network_channel.c
```diff
@@ -9,14 +9,16 @@
...
 #endif
-
 #elif defined(PLATFORM_RIOT)
 #ifdef NETWORK_CHANNEL_TCP_POSIX
 #include "platform/posix/tcp_ip_channel.c"
 #endif
 #ifdef NETWORK_CHANNEL_COAP_RIOT
 #include "platform/riot/coap_udp_ip_channel.c"
 #endif
+#ifdef NETWORK_CHANNEL_UART_RIOT
```
erlingrj, line 19: I think here we should go with `NETWORK_CHANNEL_UART` only. The reason is that in the LF program we just specify the network channel type, and not the platform. So in the generated CMakeLists/Makefile we will add compile defs for the channel type, without knowing which platform we are compiling for yet.

The same goes for COAP.

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
+#include "reactor-uc/platform/riot/uart_channel.h"
+#include "reactor-uc/logging.h"
+#include "reactor-uc/environment.h"
+#include "reactor-uc/serialization.h"
+#include "led.h"
+
+#define UART_CHANNEL_ERR(fmt, ...) LF_ERR(NET, "UARTSyncChannel: " fmt, ##__VA_ARGS__)
```
erlingrj, line 7:
```suggestion
#define UART_CHANNEL_ERROR(fmt, ...) LF_ERROR(NET, "UARTSyncChannel: " fmt, ##__VA_ARGS__)
```

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
+#include "reactor-uc/platform/riot/uart_channel.h"
+#include "reactor-uc/logging.h"
+#include "reactor-uc/environment.h"
+#include "reactor-uc/serialization.h"
+#include "led.h"
```
erlingrj, line 5: Is this needed?

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  self->state = NETWORK_CHANNEL_STATE_CONNECTED;
+}
+
+void UARTAsyncChannel_ctor(UARTAsyncChannel *self, Environment *env, uint32_t uart_device, uint32_t baud) {
+  UARTSyncChannel_ctor(&self->super, env, uart_device, baud);
+
+  cond_init(&self->receive_cv);
+  mutex_init(&self->receive_lock);
+
+  // Create decoding thread
```
erlingrj, line 155: Both "connection thread" and "decode thread" is used to describe this thread and associated variables. Wuold be good to stick to one

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+
+void UARTSyncChannel_ctor(UARTSyncChannel *self, Environment *env, uint32_t uart_device, uint32_t baud) {
+  assert(self != NULL);
+  assert(env != NULL);
+
+  self->uart_dev = UART_DEV(uart_device);
+
+  int result = uart_init(self->uart_dev, baud, _UARTSyncChannel_receive_callback, self);
+
+  if (result == -ENODEV) {
```
erlingrj, line 116: Here we should probably terminate the program with `throw()` or `validate(false)`?

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  UARTAsyncChannel *self = (UARTAsyncChannel *)arg;
+  mutex_lock(&self->receive_lock);
+
+  UART_CHANNEL_DEBUG("Entering decoding loop");
+  while (true) {
+    cond_wait(&self->receive_cv, &self->receive_lock);
+
+    UARTSyncChannel_poll((NetworkChannel *)&self->super);
+
+    thread_yield_higher();
```
erlingrj, line 102: Why this yield here? Wont it yield when arriving at `cond_wait`?

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  self->read_index++;
+
+  if (self->read_index >= minimum_message_size) {
+    if (self->super.super.category == NETWORK_CHANNEL_CATEGORY_ASYNC) {
+      cond_signal(&((UARTAsyncChannel *)self)->receive_cv);
+    }
+  }
+}
+
+void UARTSyncChannel_poll(NetworkChannel *untyped_self) {
```
erlingrj, line 73: Here we should probably read out all messages in the channel, not just the first one

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  UARTSyncChannel *self = (UARTSyncChannel *)untyped_self;
+
+  int bytes_left = deserialize_from_protobuf(&self->output, self->read_buffer, self->read_index);
```
erlingrj, line 76: here `self->read_index` is passed as the `size` argument. If this is indeed correct then I think `read_index` is not the best name for this variable. It normally points to the first unread byte. So you would read from `self->read_buffer[self->read_index]`.

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+                                                                               const FederateMessage *msg),
+                                                      FederatedConnectionBundle *conn) {
+  UART_CHANNEL_INFO("Register receive callback");
+  UARTSyncChannel *self = (UARTSyncChannel *)untyped_self;
+
+  self->receive_callback = receive_callback;
+  self->federated_connection = conn;
+}
+
+void _UARTSyncChannel_receive_callback(void *arg, uint8_t received_byte) {
```
erlingrj, line 59: It is a little confusing that this is unrelated to the other receive callback. If this is called from the ISR it might be clearer to call it interrupt_callback or something?

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  UARTSyncChannel *self = (UARTSyncChannel *)arg;
+  const uint32_t minimum_message_size = 12;
+
+  self->read_buffer[self->read_index] = received_byte;
```
erlingrj, line 63: OK I see, the `read_index` is the `write_index` of the read buffer

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,161 @@
...
+  self->read_index++;
+
+  if (self->read_index >= minimum_message_size) {
```
erlingrj, line 66: It is a little unfortunate that we will try decoding for each byte we receive. It suggests that we may want some framing around the message, i.e. a header with a length field. But that will be future work

src/platform/riot/uart_channel.c
erlingrj: IIUC `read_index` is the index to write the next byte to. I think the confusion arise from the fact that we have a `read_buffer` that we are reading and writing to. Maybe `read_index` should be `read_buffer_write_index`. Or we could rename `read_buffer` to `receive_buffer` and call it `receive_buffer_index` this way it would not be confused with reading and writing

include/reactor-uc/platform/riot/uart_channel.h
```diff
@@ -0,0 +1,46 @@
...
+struct UARTAsyncChannel {
+  UARTSyncChannel super;
+
+  char connection_thread_stack[THREAD_STACKSIZE_MAIN];
+  int connection_thread_pid;
+  mutex_t receive_lock;
+  cond_t receive_cv;
+};
+
+void UARTSyncChannel_ctor(UARTSyncChannel *self, Environment *env, uint32_t uart_device, uint32_t baud);
```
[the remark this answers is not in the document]
tanneberger: and we probably also want to specify if this a poll or async UART channel.

src/network_channel.c
tanneberger: I agree

src/platform/riot/uart_channel.c
erlingrj: This is a change I introduced recently to stay compatible with reactor-c target properties

src/platform/riot/uart_channel.c
tanneberger: Or just attempt decoding every X received bytes.

include/reactor-uc/network_channel.h
```diff
@@ -31,14 +31,27 @@ typedef enum {
 typedef enum {
   NETWORK_CHANNEL_TYPE_TCP_IP,
   NETWORK_CHANNEL_TYPE_COAP_UDP_IP,
+  NETWORK_CHANNEL_TYPE_I2C,
+  NETWORK_CHANNEL_TYPE_UART
 } NetworkChannelType;
 
+typedef enum {
+  NETWORK_CHANNEL_CATEGORY_ASYNC,
+  NETWORK_CHANNEL_CATEGORY_SYNC,
```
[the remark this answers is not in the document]
tanneberger: done

include/reactor-uc/network_channel.h
tanneberger: done

src/platform/riot/uart_channel.c
tanneberger: renamed it to `decoding_thread`

src/platform/riot/uart_channel.c
tanneberger: done

src/platform/riot/uart_channel.c
tanneberger: done

src/platform/riot/uart_channel.c
tanneberger: its now `receive_buffer` and `receive_buffer_index`

src/platform/riot/uart_channel.c
tanneberger: done

src/platform/riot/uart_channel.c
tanneberger: I think yes

src/platform/riot/uart_channel.c
erlingrj: Yes, but that could lead to unexpected results if you receive messages very rarely. Imagine you have received the entire message, but you have received X-1 bytes since the last decode attempt. So we would also need to take time into account.

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,163 @@
...
+    if (bytes_left >= 0) {
+      int receive_buffer_index = self->receive_buffer_index;
+      self->receive_buffer_index = bytes_left;
+      memcpy(self->receive_buffer, self->receive_buffer + (receive_buffer_index - bytes_left), bytes_left);
+
+      if (self->receive_callback != NULL) {
+        UART_CHANNEL_DEBUG("calling user callback!");
+        self->receive_callback(self->federated_connection, &self->output);
+      }
+    }
```
erlingrj, line 89: In the scenario that we have received some bytes (more than 12), but we have not received enough that we can parse an entire message, then I think the right thing is to break out, rather than wait until the message is entirely received

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,163 @@
...
+  const uint32_t minimum_message_size = 12;
+
+  while (self->receive_buffer_index > minimum_message_size) {
+    int bytes_left = deserialize_from_protobuf(&self->output, self->receive_buffer, self->receive_buffer_index);
+    UART_CHANNEL_DEBUG("Bytes Left after attempted to deserialize %d", bytes_left);
+
+    if (bytes_left >= 0) {
+      int receive_buffer_index = self->receive_buffer_index;
+      self->receive_buffer_index = bytes_left;
+      memcpy(self->receive_buffer, self->receive_buffer + (receive_buffer_index - bytes_left), bytes_left);
```
erlingrj, line 83: For this to work, `UartPollChannel_poll` must always be called within a critical section, so that interrupts are disabled. A few other places I added `_locked` postfix to functions that has to be called from critical sections. Maybe this would be wise for this one as well?

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,163 @@
...
+
+void *_UARTAsyncChannel_decode_loop(void *arg) {
+  UARTAsyncChannel *self = (UARTAsyncChannel *)arg;
+  mutex_lock(&self->receive_lock);
+
+  UART_CHANNEL_DEBUG("Entering decoding loop");
+  while (true) {
+    cond_wait(&self->receive_cv, &self->receive_lock);
+
+    UARTPollChannel_poll((NetworkChannel *)&self->super);
```
erlingrj, line 101: I think we have to disable IRQs before calling `UARTPollChannel_poll` here. Even if you hold the `receive_lock`, I dont think there is anything stopping an interrupt from arriving and start messing around with the receive buffer while you are in the middle of a copy.

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,163 @@
...
+    throw("The user specified an invalid UART device!");
+  } else if (result == ENOTSUP) {
+    UART_CHANNEL_ERR("Given configuration to uart is not supported!");
+    throw("The given combination of parameters for creating a uart device is unsupported!");
+  } else if (result < 0) {
+    UART_CHANNEL_ERR("UART Init error occurred");
+    throw("Unknown UART RIOT Error!");
+  }
+
+  result = uart_mode(self->uart_dev, UART_DATA_BITS_8, UART_PARITY_EVEN, UART_STOP_BITS_2);
```
erlingrj, line 126: These options should be possible to set from the lf file, I think something like:

```
@interface_uart(device_id=1, baudrate=115200, data_bits=8, parity="none"/"even"/"odd", stop_bits=1)
f = new UartFed()
```

I.e. they must be exposed through the UartPollChannel_ctor. But we shouldnt expose the RIOT specific types, because we want to provide implementations for Zephyr and Posix also. So you might need to create your own typedefs for this.

Also, the normal default is 8N1 (8 data bits, no parity bit, and 1 stop bit)

erlingrj (changes requested):
This is starting to look real good, nice progress. I pushed a few trivial fixes, but I think there still is some race condition between the ISR and the runtime. The other day we discussed lock-free queues, but the current implementation needs lock (or disabling interrupts). We could make it lock free by implementing a circular buffer between the ISR and the runtime (but we would then have to move the data from this circular buffer to a "linear" buffer before feeding it forward to the deserialization pipeline. 

Also, I realized that we should expose more UART settings to the user. 

Keep up the good work!!

src/platform/riot/uart_channel.c
tanneberger: also true

src/platform/riot/uart_channel.c
tanneberger: I would suggest we protect updates to th `receive_buffer_index` but the `receive_buffer` I would not protect.

src/platform/riot/uart_channel.c
tanneberger: done, there are now integer parameters to the constructors.

tanneberger: done

src/platform/riot/uart_channel.c
tanneberger: incrementing overwriting the `receive_buffer_index` is now protected.

src/platform/riot/uart_channel.c
tanneberger: lets change this when we converged in the discussion #199

src/platform/riot/uart_channel.c
erlingrj: Hmm, I do not think this is safe. What happens e.g. if the interrupt occurs between line 79 and 84? Say we have parsed out a message, and that there are 4 bytes left. Well, if the interrupt occurs, then really there should be 5 bytes left, so when we do the memcpy to copy the bytes back to the beginning of the buffer, we forget the fifth byte and it is lost.

I think the memcpy has to be part of the critical section as well

src/platform/riot/uart_channel.c
```diff
@@ -0,0 +1,231 @@
...
+    riot_data_bits = UART_DATA_BITS_8;
+    break;
+  default:
+    UART_CHANNEL_ERR("Invalid data_bits given to UARTPollChannel %u", data_bits);
+    throw("Invalid data_bits given!");
+  };
+
+  uart_parity_t riot_parity;
+
+  switch (parity) {
```
erlingrj, line 157: Not to be a nitpicker, but wouldn't it be nicer to define our own enum for this. And while we are at it, we could define it for data_bits and stop_bits also?

src/platform/riot/uart_channel.c
erlingrj: What is the LED used for, I cant see any code that uses something LED related?

src/platform/riot/uart_channel.c
tanneberger: I agree I went for integers because if we define an enum that this enum is not platform specific and should not be definded inside platform specific code.

tanneberger: Where should this enum go? 

- platform
   - riot
      - uart_channel.h
- network_channel
    - tcp_ip_channel.h
   - uart_channel.h <-- inside here

tanneberger:
@erlingrj there is not `include/reactor-uc/network_channel/uart_channel.h` for the platform independent uart enums.

src/platform/riot/uart_channel.c
tanneberger: What I would prefer to this is turning the `receive_buffer_index` into an atomic, then we could get rid of that lock.

I think only protecting the `receive_buffer_index` is enough, because the interrupts are sequential so the incrementing is sequential and overwriting the `receive_buffer_index` also gets sequenialized because we lock this region.

lfc/core/src/main/java/org/lflang/validation/AttributeSpec.java
```diff
@@ -253,6 +253,19 @@ enum AttrParamType {
...
     // address="127.0.0.1")
+
+    ATTRIBUTE_SPECS_BY_NAME.put(
+        "interface_uart",
+        new AttributeSpec(
+            List.of(
+                new AttrParamSpec("name", AttrParamType.STRING, true),
+                new AttrParamSpec("uart_device", AttrParamType.INT, true),
+                new AttrParamSpec("baud_rate", AttrParamType.INT, true),
+                new AttrParamSpec("data_bits", AttrParamType.STRING, true),
```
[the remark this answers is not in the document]
tanneberger: yes, STRING is also in caps.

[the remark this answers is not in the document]
erlingrj: What I meant was: shouldnt the data_bits attribute parameter be an integer instead of a string?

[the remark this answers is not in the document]
erlingrj: E.g. `8` instead of `"8"`

erlingrj:
I think we should add at least one test LF program before merging this, showing how the annotation looks. If you put it under `test/lf/src/only_build` and set `platform: RIOT`, we will only run code-gen on it. 

As noted above, I think `data_bits` and `stop_bits` attribute parameters should be integers, not strings

lfc/core/src/main/java/org/lflang/validation/AttributeSpec.java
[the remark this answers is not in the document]
tanneberger: done

include/reactor-uc/network_channel.h
```diff
@@ -31,14 +31,26 @@ typedef enum {
 typedef enum {
   NETWORK_CHANNEL_TYPE_TCP_IP,
   NETWORK_CHANNEL_TYPE_COAP_UDP_IP,
+  NETWORK_CHANNEL_TYPE_UART
 } NetworkChannelType;
 
+typedef enum {
+  NETWORK_CHANNEL_MODE_ASYNC,
+  NETWORK_CHANNEL_MODE_POLL,
+} NetworkChannelMode;
```
LasseRosenow, line 40: Maybe call it `NetworkChannelExecutionMode` to be more clear what it is for

include/reactor-uc/network_channel.h
```diff
@@ -95,6 +107,19 @@ struct NetworkChannel {
   void (*free)(NetworkChannel *self);
 };
 
+struct SyncNetworkChannel {
```
LasseRosenow, line 110: The `NetworkChannelMode` enum is either `poll` or `async`, but the `NetworkChannel` subtypes are `SyncNetworkChannel` or `AsyncNetworkChannel`

Maybe it makes sense to align these and either have:
`NetworkChannelMode` = `Async` | `Sync`

or have:

`SyncNetworkChannel` and `PollNetworkChannel`

include/reactor-uc/network_channel.h
```diff
@@ -95,6 +107,19 @@ struct NetworkChannel {
...
+  NetworkChannel super;
+
+  /**
+   * @brief Polls for new data and calls the callback handler if a message is successfully decoded
+   */
+  void (*poll)(NetworkChannel *self);
+};
+
+struct AsyncNetworkChannel {
+  NetworkChannel super;
```
LasseRosenow, line 120: Does it maybe make sense to move the async callback out of `NetworkChannel` and into `AsyncNetworkChannel` instead of only inheriting it?

erlingrj: No, because it is also used by the PolledNetworkChannel

LasseRosenow (approved):
This looks very nice, I left some comments mostly about naming things, so nothing important.

include/reactor-uc/network_channel.h
erlingrj: Hmm, I think mode is OK for now, if we need to add more properties then we should consider being more specific

PR #192, opened 2025-01-20
LFC federated RIOT fixes

LasseRosenow:
## Build:
```bash
FEDERATION=r1 make all
```
or
```bash
FEDERATION=r2 make all
```

## Or directly Build and Run:
```bash
FEDERATION=r1 make all term
```
or
```bash
FEDERATION=r2 make all term
```


lfc/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
```diff
@@ -242,16 +242,24 @@ class UcTcpIpChannel(
...
+        }
+    }
 
     override fun generateChannelCtorSrc() =
-        "CoapUdpIpChannel_ctor(&self->channel, parent->env, \"${if (serverLhs) srcAddr else destAddr}\", AF_INET, ${serverLhs});"
+        "CoapUdpIpChannel_ctor(&self->channel, parent->env, \"${destAddr.ipAddress.address}\", ${getIpProtocolFamily(destAddr.ipAddress)});"
 
     override fun generateChannelCtorDest() =
-        "CoapUdpIpChannel_ctor(&self->channel, parent->env, \"${if (serverLhs) srcAddr else destAddr}\" AF_INET, ${!serverLhs});"
+        "CoapUdpIpChannel_ctor(&self->channel, parent->env, \"${srcAddr.ipAddress.address}\", ${getIpProtocolFamily(srcAddr.ipAddress)});"
```
LasseRosenow, line 262: Not 100% sure yet if I did this correct, but we don't have client and server, so we always want to have the other federates ip in the constructor.

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
```diff
@@ -242,17 +242,24 @@ class UcTcpIpChannel(
 class UcCoapUdpIpChannel(
```
tanneberger, line 242: Is this homogeneous enough that one code-generator class works potentially for all platforms?

erlingrj: We will find out when we add support for native and Zephyr

examples/riot/coap_federated_lf/Makefile
```diff
@@ -0,0 +1,35 @@
+REACTOR_UC_PATH ?= $(CURDIR)/../../../
+
+# The name of the LF application inside "./src" to build/run/flash etc.
+LF_MAIN ?= CoapFederatedLF
+FEDERATE ?= r1
```
erlingrj, line 5: @LasseRosenow , I took the liberty to rename this from FEDERATION to FEDERATE. The entire distributed program is a federation, while each node is a federate

examples/riot/hello_lf/Makefile
```diff
@@ -8,7 +8,7 @@ LF_MAIN ?= HelloLF
 # CFLAGS += -DNETWORK_CHANNEL_COAP_RIOT
 
 # Execute the LF compiler if build target is "all"
-ifeq ($(MAKECMDGOALS),all)
+ifeq ($(firstword $(MAKECMDGOALS)),all)
```
erlingrj, line 11: Out of curiosity, why is it necessary for `all` to be the first target?

examples/riot/coap_federated/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+REMOTE_ADDRESS=fe80::8cc3:33ff:febb:1b3 make all -C ./sender
+REMOTE_ADDRESS=fe80::8cc3:33ff:febb:1b3 make all -C ./receiver
```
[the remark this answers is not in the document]
erlingrj: Sounds reasonable. We must find a good way to build federated programs. Your current proposal, with a single Makefile, works now. But I think in the future we want independent projects for each federate because there might several difference between the boards, and their configuration.

Two questions:
1. Are you now able to pick a IP6 address from the tap interface?
2. Why do they both specify the same remote?

erlingrj (changes requested):
Looks good, but I wasnt able to run the coap_federated_lf on my end

examples/riot/coap_federated_lf/Makefile
LasseRosenow: Thank you :)

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
erlingrj: looks good to me

examples/riot/hello_lf/Makefile
LasseRosenow: Because they are executed in order and we first need to build the project and then run term.
But actually the fix is not so much about that. The main issue was, that before it only worked if the argument is `all`. So if anything came after `all` it would not work.

This way it is more relaxed and only checks if `all` comes first and doesn't care what comes after ;)

LasseRosenow:
> Are you now able to pick a IP6 address from the tap interface?

Not yet. That will come later the `build.sh` is only to let the CI build the example

LasseRosenow:
> Why do they both specify the same remote

Because the remote doesn't matter just for building the project, but I have updated the `build.sh` files to make them more `realistic` to how a real scenario would look like :)

This all will be better when I have a reliable way to extract the IP addresses, but my focus was on getting the tests run reliably right now, so I couldn't look into it yet

LasseRosenow:
> Your current proposal, with a single Makefile, works now. But I think in the future we want independent projects for each federate because there might several difference between the boards, and their configuration.

This PR is only about fixing the riot makefile integration to be able to build. The structure of the example is without any indication or proposal in that regard :) This is just the bare minimal that is needed to let it compile and find the makefiles in the correct `federate` directories in `src-gen`. 

Once we have agreed on a solution how to structure it, I can align it  :)

PR #195, opened 2025-01-23
Add lfc.cmake

erlingrj:
This PR adds a cmake file that defines three functions that are useful for cmake-based build systems, such as Zephyr/west.

Here is an example using `lfc.cmake`:
https://github.com/lf-lang/lf-zephyr-uc-template/blob/d2f00313c58a76f15e40dee08efb545dd21b7bca/CMakeLists.txt

### DIFF
--- a/cmake/lfc.cmake
+++ b/cmake/lfc.cmake
@@ -1,0 +1,69 @@
+cmake_minimum_required(VERSION 3.20.0)
+
+# Sets up default values for LF_MAIN and LOG_LEVEL which
+# can be overridden by the user, either in the main CMakeLists.txt or
+# from the command line.
+function (lf_setup)
+  if (NOT DEFINED LF_MAIN)
+    message(FATAL_ERROR "LF_MAIN must be defined to be the name of the main LF source file without the .lf extension")
+  endif()
+
+  if (NOT DEFINED LOG_LEVEL)
+    set(LOG_LEVEL "LF_LOG_LEVEL_INFO" PARENT_SCOPE)
+  endif()
+endfunction()
+
+# Run the LFC compiler on the specified LF source file, LF_SOURCE_DIR/LF_MAIN.lf. Also make the CMake configuration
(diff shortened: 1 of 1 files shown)


LasseRosenow (approved):
My mind gets twisted reading so much cmake 😵‍💫, but it looks very good to me.
I like how small and concise the `user` file has become :)

Nice work! 💪

erlingrj:
> My mind gets twisted reading so much cmake 😵‍💫, but it looks very good to me. I like how small and concise the `user` file has become :)
> 
> A little bit jealous of the function feature :O
> 
> Nice work! 💪

Yes, CMake is terrible

LasseRosenow:
One small question: In the RIOT make based solution I had the issue that just specifying `LF_MAIN` didn't work with federated programs, because in that case the path is more complex and the generated code is not under `src-gen/${LF_MAIN}/Makefile` anymore, but under `src-gen/${LF_MAIN}/r1/Makefile`.

Is the idea in this case to write something like this:
`lf_build_generated_code(app ${CMAKE_CURRENT_SOURCE_DIR}/src-gen/${LF_MAIN}/r1)`
`

tanneberger (approved):
Looks fine to me.

erlingrj:
> One small question: In the RIOT make based solution I had the issue that just specifying `LF_MAIN` didn't work with federated programs, because in that case the path is more complex and the generated code is not under `src-gen/${LF_MAIN}/Makefile` anymore, but under `src-gen/${LF_MAIN}/r1/Makefile`. That is why I introduced the `FEDERATE` variable there.
> 
> Is the idea in this case to write something like this: `lf_build_generated_code(app ${CMAKE_CURRENT_SOURCE_DIR}/src-gen/${LF_MAIN}/r1)` ?

Yes exactly! In the federated zephyr CMake template I have a `FEDERATE` variable and we use this to find the correct src-gen path

PR #198, opened 2025-01-23
Attributes for action and connection buffers

erlingrj:


lfc/core/src/main/java/org/lflang/AttributeUtils.java
```diff
@@ -290,6 +291,24 @@ public static List<Attribute> getInterfaceAttributes(Instantiation node) {
     return findAttributesByNameStartingWith(node, "interface");
   }
 
+  public static int getMaxNumberOfPendingEvents(Action node) {
+    Attribute attr =findAttributeByName(node, "max_pending_events");
```
tanneberger, line 295: We should add formatting for the java code.

tanneberger (changes requested):

lfc/core/src/main/java/org/lflang/AttributeUtils.java
erlingrj: Thanks, yes. There is a setup in `lingua-franca` with spotless, maybe I can just copy it

PR #201, opened 2025-01-24
Run spotless on LFC and add check to CI

erlingrj:
This just uses the Spotless formatter already part of LFC, also adds a check to CI. To run the formatted simply:
`cd lfc && ./gradlew spotlessApply`


erlingrj:
FYI @LasseRosenow and @tanneberger

PR #206, opened 2025-01-28
Add maxwait, start time coordination and system events

erlingrj:
This PR took on a lot of weight unfortunately and now contains three major features:

1. The `maxwait` property.
This PR adds a `maxwait` property for each reaction with an optional STP violation handler. This includes changes the the code-generator and the grammar. In particular we must find the maximum `maxwait` associated with any reaction triggered by an input. We also add an `intended_tag` field to events so that we can detect the STP violations.

2. Startup coordination.
We also add a StartupCoordinator which each federate will have one of. It does a blocking gossip algorithm at boot negotiating the startup tag of the federation. 

3. SystemEvents
Finally, this PR adds the notion of system events and a system event queue. Still this is not used, but I will use it for implementing the ClockSync algorithm. The idea is that any received coordination message will be put on the SystemEventQueue so that we can do blocking IO in response to received messages.


external/proto/message.pb.h
```diff
@@ -10,34 +10,64 @@
...
-    MessageType_TAGGED_MESSAGE = 1,
-    MessageType_START_TAG_SIGNAL = 2
-} MessageType;
+typedef enum _StartupCoordinationState {
+    StartupCoordinationState_UNINITIALIZED = 0,
+    StartupCoordinationState_CONNECTING = 1,
+    StartupCoordinationState_HANDSHAKING = 2,
+    StartupCoordinationState_NEGOTIATING = 3,
+    StartupCoordinationState_RUNNING = 4
+} StartupCoordinationState;
```
LasseRosenow, line 19: Is the connecting step here the waiting for the network channel to connect?

include/reactor-uc/tag.h
```diff
@@ -71,7 +75,7 @@ typedef int64_t interval_t;
 /**
  * Microstep instant.
  */
-typedef uint32_t microstep_t;
+typedef int32_t microstep_t;
```
LasseRosenow, line 78: A microstep can be negative?

src/startup_coordinator.c
```diff
@@ -0,0 +1,269 @@
...
+    self->msg.message.startup_coordination.message.startup_handshake_response.state = self->state;
+    self->neighbor_state[idx].handshake_response_sent = true;
+    self->env->leave_critical_section(self->env);
+    chan->send_blocking(chan, &self->msg);
+    self->env->enter_critical_section(self->env);
+  }
+}
+
+/**
+ * @brief Perform a handshake with all neighbors.abort
```
LasseRosenow, line 114: with all `neighbors.abort`?

erlingrj: Typo, thanks!

include/reactor-uc/tag.h
erlingrj: No, thanks!

erlingrj:
Thanks for the review. The segfault caught by RIOT was a mistake by me. Running tests on many different platforms turns out to be an asset as we are more likely to catch memory bugs

PR #208, opened 2025-01-31
Generate fed templates

erlingrj:
Based on our discussions https://github.com/lf-lang/reactor-uc/discussions/203 I have started this draft PR.

By calling `lfc --gen-fed-templates src/Hello.LF` we generate project templates for each federate based on the platform. A platform is either inferred from the top-level target property (`platform: Native`) or from an attribute, currently:
```
@platform_riot
f1 = new Fed()

@platform_zephyr
f2 = new Fed()
```

Currently still a draft. @LasseRosenow, could you help me out a little with the RIOT makefile? Check out the file UcFederateTemplateProjectGenerator.kt, currently it is an extremely simple generator for the different files. We can make a nicer design if this gets more complicated


LasseRosenow:
Nice, I will have a look when I have time later

Issue #212, opened 2025-02-04
Run basic tests on macOS also.


erlingrj:
If we commit to supporting running on macOS using `platform: Native` then we should run the tests in CI

Issue #211, opened 2025-02-03
Eventfd is not supported on MacOs.


tanneberger:
MacOs user trying to compile reactor-uc will get the following compile error:

```
./src/platform/posix/tcp_ip_channel.c:15:10: fatal error: 'sys/eventfd.h' file not found
   15 | #include <sys/eventfd.h>
```

Is there a posix complaint alternative?

erlingrj:
We already had this issue. And it was solved here: https://github.com/lf-lang/reactor-uc/pull/178

Did we forget to remove the include statement?

PR #214, opened 2025-02-04
Remove unused eventfd and add macOS testing to CI

erlingrj:
Addresses #212 and #211


erlingrj:
Hm, seems to be some issues with macOS and federated to look into

PR #218, opened 2025-02-07
ktfmt

tanneberger:
adding ktfmt. 

I added the kotlin default style, but I am also happy to use the other styles.


erlingrj (approved):
Great, thank you!

PR #221, opened 2025-02-10
Remove platform from generated NETWORK_CHANNEL_X CFLAG

LasseRosenow:
We renamed `NETWORK_CHANNEL_COAP_RIOT` etc. to `NETWORK_CHANNEL_COAP` etc. 
This updates the compiler to also comply with this change and makes the makefile RIOT independent again 🥳


erlingrj (approved):
Great, thanks!

PR #228, opened 2025-02-18
Add clock synchronization between federates

erlingrj:
This PR adds clock synchronization between the federates. It does so using the new system event infrastructure. It does several things:

1. Updates the start tag negotiation. Now only clock grand masters (GM) can propose a start time. Other federates start by proposing NEVER. When the start time have converged, the other federates performs an initial sync, setting their current local clock to the start time. This will be a rough estimate. But we need some initial sync for startup, imagine if one federate thinks it is 2008 and the other 2040.

2. A priority distribution scheme. GMs starts by broadcasting their priority which will trigger the others to also broadcast their priority. Every federate that is not a grandmaster will start synchronizing to its neighbor with the lowest priority (least steps away from a GM). Every federate can act as a local master to its neighbor responding to clock sync messages.

3. The simple PTP algorithm with four timestamps and a PI controller is implemented. This will be improved upon with coded probes when I have it working. In the future we can also consider using the "network effect" idea from the coded probes paper. Basically, compute clock offset to all your neighbors and then compute the transitive offset in the cycles of the federation, they should sum up to 0. But for now we go with the simple PTP implementation to have a working base.

4. A PhysicalClock is implemented as a wrapper around the platform clock. The program calls `PhysicalClock::get_time` to read the current time. We can also set the time and apply a frequency adjustment to the clock. It calls `Platform::get_physical_time` under the hood and applies drift correction etc.

TODO:
- [x] Add support for configuring the clock-sync parameters (especially interval, max -adjustment maybe and grandmaster)
- [x] Verify that the federates make up a connected graph.


src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+static void ClockSynchronization_schedule_system_event(ClockSynchronization *self, instant_t time, int message_type) {
+  ClockSyncEvent *payload = NULL;
+  lf_ret_t ret;
+  ret = self->super.payload_pool.allocate(&self->super.payload_pool, (void **)&payload);
+  if (ret != LF_OK) {
+    LF_ERR(CLOCK_SYNC, "Failed to allocate payload for clock-sync system event.");
+    validate(false);
+    return;
+  }
+  payload->neighbor_index = -1; // -1 means that it is from ourself
```
erlingrj, line 47: Magic number

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  payload->msg.which_message = message_type;
+  tag_t tag = {.time = time, .microstep = 0};
+  SystemEvent event = SYSTEM_EVENT_INIT(tag, &self->super, (void *)payload);
+  // TODO: Critical section?
+
+  ret = self->env->scheduler->schedule_at_locked(self->env->scheduler, &event.super);
+  if (ret != LF_OK) {
+    LF_ERR(CLOCK_SYNC, "Failed to schedule clock-sync system event.");
+    validate(false);
```
erlingrj, line 56: Free payload?

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  }
+}
+
+static int ClockSynchronization_my_priority(ClockSynchronization *self) {
```
erlingrj, line 60: Document

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  if (self->is_grandmaster) {
+    return 0;
```
erlingrj, line 62: GRANDMASTER_PRIORITY?

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  for (size_t i = 0; i < self->num_neighbours; i++) {
+    int neighbor_priority = self->neighbor_clock[i].priority;
+    if (neighbor_priority != UNKNOWN_PRIORITY && (neighbor_priority + 1) < my_priority) {
+      my_priority = neighbor_priority + 1;
+    }
+  }
+  return my_priority;
+}
+
+static void ClockSynchronization_handle_message_callback(ClockSynchronization *self, const ClockSyncMessage *msg,
```
erlingrj, line 75: Document

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+                                                         size_t bundle_idx) {
+  (void)self;
+  (void)msg;
+  (void)bundle_idx;
```
erlingrj, line 79: Unnecessary?

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  lf_ret_t ret = self->super.payload_pool.allocate_with_reserved(&self->super.payload_pool, (void **)&payload,
+                                                                 NUM_RESERVED_EVENTS);
+  if (ret == LF_OK) {
+    payload->neighbor_index = bundle_idx;
+    memcpy(&payload->msg, msg, sizeof(ClockSyncMessage));
+    tag_t now = {.time = self->env->get_physical_time(self->env), .microstep = 0};
+    SystemEvent event = SYSTEM_EVENT_INIT(now, &self->super, (void *)payload);
+    ret = self->env->scheduler->schedule_at_locked(self->env->scheduler, &event.super);
+    if (ret != LF_OK) {
+      LF_ERR(CLOCK_SYNC, "Failed to schedule clock-sync system event.");
```
erlingrj, line 91: free payload?

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+      assert(false);
+    }
+  } else {
+    LF_WARN(CLOCK_SYNC, "Failed to allocate payload for clock-sync system event.");
+  }
+}
+
+static void ClockSynchronization_handle_priority_request(ClockSynchronization *self, int src_neighbor) {
+  int my_priority = ClockSynchronization_my_priority(self);
```
erlingrj, line 100: Reoroder

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  lf_ret_t ret;
+
+  if (src_neighbor == -1) {
```
erlingrj, line 103: magic number

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+    LF_DEBUG(CLOCK_SYNC, "Broadcasting priority %d to all neighbors", my_priority);
+    for (size_t i = 0; i < self->num_neighbours; i++) {
+      if (((int)i) == self->master_neighbor_index) {
```
erlingrj, line 106: Doc

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  int my_new_priority = ClockSynchronization_my_priority(self);
+  if (my_new_priority < my_old_priority) {
+    LF_DEBUG(CLOCK_SYNC, "Received priority %d+1=%d less than current priority %d, updating", priority, my_new_priority,
+             my_old_priority);
+    self->master_neighbor_index = src_neighbor;
+    ClockSynchronization_handle_priority_request(self, -1);
+  }
+}
+
+static void ClockSynchronization_handle_delay_request(ClockSynchronization *self, SystemEvent *event) {
```
erlingrj, line 144: Docs

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  bundle->send_msg.message.clock_sync_msg.message.sync_response.time = event->super.tag.time;
+  bundle->send_msg.message.clock_sync_msg.message.sync_response.sequence_number =
+      payload->msg.message.delay_request.sequence_number;
+  ret = chan->send_blocking(chan, &bundle->send_msg);
+  if (ret != LF_OK) {
+    LF_WARN(CLOCK_SYNC, "Failed to send DelayResponse to neighbor %zu", src_neighbor);
+  }
+}
+
+static void ClockSynchronization_handle_sync_response(ClockSynchronization *self, SystemEvent *event) {
```
erlingrj, line 164: Docs

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  self->timestamps.t1 = msg->time;
+  self->timestamps.t2 = event->super.tag.time;
+
+  FederatedConnectionBundle *bundle = self->env->net_bundles[src_neighbor];
+  NetworkChannel *chan = bundle->net_channel;
+  bundle->send_msg.which_message = FederateMessage_clock_sync_msg_tag;
+  bundle->send_msg.message.clock_sync_msg.which_message = ClockSyncMessage_delay_request_tag;
+  bundle->send_msg.message.clock_sync_msg.message.delay_request.sequence_number = ++self->sequence_number;
```
erlingrj, line 184: Should we increment it here? Makes sense to have the same sequence number for entire round?

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+  }
+  self->timestamps.t4 = msg->time;
+  ClockSynchronization_correct_clock(self, &self->timestamps);
+}
+
+static void ClockSynchronization_handle_request_sync(ClockSynchronization *self, SystemEvent *event) {
+  ClockSyncEvent *payload = (ClockSyncEvent *)event->super.payload;
+  lf_ret_t ret;
+  int src_neighbor = payload->neighbor_index;
+  if (src_neighbor == -1) {
```
erlingrj, line 208: Magic number

src/clock_synchronization.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+void ClockSynchronization_ctor(ClockSynchronization *self, Environment *env, NeighborClock *neighbor_clock,
+                               size_t num_neighbors, bool is_grandmaster, size_t payload_size, void *payload_buf,
+                               bool *payload_used_buf, size_t payload_buf_capacity) {
+  self->env = env;
+  self->neighbor_clock = neighbor_clock;
+  self->num_neighbours = num_neighbors;
+  self->is_grandmaster = is_grandmaster;
+  self->master_neighbor_index = -1;
```
erlingrj, line 283: Magic number

src/physical_clock.c
```diff
@@ -0,0 +1,58 @@
+#include "reactor-uc/physical_clock.h"
+#include "reactor-uc/logging.h"
+
+// FIXME: Think about how to handle the case where the clock is stepped back.
```
erlingrj, line 4: Is that a problem?

src/physical_clock.c
```diff
@@ -0,0 +1,58 @@
...
+// FIXME: How to handle the case where the clock is stepped forward.
```
erlingrj, line 5: Defintely not a problem

src/physical_clock.c
```diff
@@ -0,0 +1,58 @@
...
+// FIXME: How to handle concurrency?
```
erlingrj, line 6: Document this well. The only problem is between async context calling `get_time` and runtime context calling `set_time` or `adjust_time`. So it means that we should only call these things from critical sections.

src/physical_clock.c
```diff
@@ -0,0 +1,58 @@
...
+
+lf_ret_t PhysicalClock_set_time(PhysicalClock *self, instant_t time) {
```
erlingrj, line 8: Document all functions

erlingrj (commented):
Self-review

PR #230, opened 2025-02-21
UartChannel for Raspberry Pi Pico

tanneberger:


include/reactor-uc/platform/pico/uart_channel.h
```diff
@@ -0,0 +1,39 @@
...
+
+typedef struct FederatedConnectionBundle FederatedConnectionBundle;
+typedef struct UartPolledChannel UartPolledChannel;
+typedef struct UartAsyncChannel UartAsyncChannel;
+
+#define UART_CHANNEL_BUFFERSIZE 1024
+// The UartChannel is not connection-oriented and will always appear as connected, so no need to wait.
+#define UART_CHANNEL_EXPECTED_CONNECT_DURATION SEC(5)
+
+extern UartPolledChannel *uart_channel_0;
```
erlingrj, line 20: Why are these here in the header file? Are they used anywhere outside `uart_channel.c`

include/reactor-uc/platform/pico/uart_channel.h
```diff
@@ -0,0 +1,39 @@
...
+#include "hardware/uart.h"
+#include "hardware/irq.h"
+
+typedef struct FederatedConnectionBundle FederatedConnectionBundle;
+typedef struct UartPolledChannel UartPolledChannel;
+typedef struct UartAsyncChannel UartAsyncChannel;
+
+#define UART_CHANNEL_BUFFERSIZE 1024
+// The UartChannel is not connection-oriented and will always appear as connected, so no need to wait.
+#define UART_CHANNEL_EXPECTED_CONNECT_DURATION SEC(5)
```
erlingrj, line 18: 5 seconds? This tells the runtime to sleep 5 seconds between each connection attempt

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+#include "reactor-uc/serialization.h"
+
+#define UART_CHANNEL_ERR(fmt, ...) LF_ERR(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_WARN(fmt, ...) LF_WARN(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_INFO(fmt, ...) LF_INFO(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_DEBUG(fmt, ...) LF_DEBUG(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+
+extern Environment *_lf_environment;
+
+UartPolledChannel *uart_channel_0 = NULL;
```
erlingrj, line 13: Couldn't these, and the FederateMessage be on the UartPollchannel struct instead as global variables?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+#include "reactor-uc/platform/pico/pico.h"
+#include "reactor-uc/logging.h"
+#include "reactor-uc/serialization.h"
+
+#define UART_CHANNEL_ERR(fmt, ...) LF_ERR(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_WARN(fmt, ...) LF_WARN(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_INFO(fmt, ...) LF_INFO(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+#define UART_CHANNEL_DEBUG(fmt, ...) LF_DEBUG(NET, "UartPolledChannel: " fmt, ##__VA_ARGS__)
+
+extern Environment *_lf_environment;
```
erlingrj, line 11: I believe you will get this declaration by doing `#include "reactor-uc/environment.h"`

src/platform/pico/uart_channel.c
erlingrj: If they have to be global, lets make them static at least

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+
+UartPolledChannel *uart_channel_0 = NULL;
+UartPolledChannel *uart_channel_1 = NULL;
+FederateMessage msg;
+
+static lf_ret_t UartPolledChannel_open_connection(NetworkChannel *untyped_self) {
+  UartPolledChannel *self = (UartPolledChannel *)untyped_self;
+  char connect_message[] = {0x7, 0x7, 0x7, 0x7};
```
erlingrj, line 19: Probably smart to make {0x7 .... } a compile def instead of magic numbers

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  uart_write_blocking(self->uart_device, (const uint8_t *)connect_message, 4);
+  UART_CHANNEL_DEBUG("Open connection");
+  return LF_OK;
+}
+
+static void UartPolledChannel_close_connection(NetworkChannel *untyped_self) {
```
erlingrj, line 25: Perhaps we should send another message here to signal that it is closed to the other side?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  UART_CHANNEL_DEBUG("Close connection");
+  (void)untyped_self;
+}
+
+static void UartPolledChannel_free(NetworkChannel *untyped_self) {
+  UART_CHANNEL_DEBUG("Free");
+  (void)untyped_self;
+}
+
+static bool UartPolledChannel_is_connected(NetworkChannel *untyped_self) {
```
erlingrj, line 35: This one just sleeps for 500 msec before saying that we are connected?

erlingrj: This function was intended to check for a connection and possibly do some work towards establishing the connection and returning. We expect the runtime to poll this function for all channels, one after the other, and then sleeping, until they all return true

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  lf_ret_t wait_return = LF_SLEEP_INTERRUPTED;
+  while (wait_return == LF_SLEEP_INTERRUPTED) {
+    wait_return = _lf_environment->wait_until(_lf_environment, wake_up_time);
+  }
+  return self->state == NETWORK_CHANNEL_STATE_CONNECTED;
+}
+
+static lf_ret_t UartPolledChannel_send_blocking(NetworkChannel *untyped_self, const FederateMessage *message) {
+  UartPolledChannel *self = (UartPolledChannel *)untyped_self;
+  int message_size = serialize_to_protobuf(message, self->send_buffer, UART_CHANNEL_BUFFERSIZE);
```
erlingrj, line 49: Check return value

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+                                                                                 const FederateMessage *message),
+                                                        FederatedConnectionBundle *bundle) {
+  UART_CHANNEL_DEBUG("Register receive callback");
+  UartPolledChannel *self = (UartPolledChannel *)untyped_self;
+
+  self->receive_callback = receive_callback;
+  self->bundle = bundle;
+}
+
+void _UartPolledChannel_interrupt_handler(UartPolledChannel *self) {
```
erlingrj, line 66: static?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  while (uart_is_readable(self->uart_device)) {
+    uint8_t received_byte = uart_getc(self->uart_device);
+    self->receive_buffer[self->receive_buffer_index] = received_byte;
+    self->receive_buffer_index++;
```
erlingrj, line 70: We should check for overflow, that we don't increment receive_buffer_index beyond the max size.

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  }
+  if (self->receive_buffer_index >= 30) {
```
erlingrj, line 72: 30 should be a a compile def. And several messages are less than thirty. IIRC the smallest message is 6-7 bytes only.

erlingrj: So for safety we should set this to 5, and solve it with framing

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+    _lf_environment->platform->new_async_event(_lf_environment->platform);
+  }
+}
+
+void _UartPolledChannel_pico_interrupt_handler_0(void) {
```
erlingrj, line 77: static?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  if (uart_channel_0 != NULL) {
+    _UartPolledChannel_interrupt_handler(uart_channel_0);
+  }
+}
+
+void _UartPolledChannel_pico_interrupt_handler_1(void) {
```
erlingrj, line 83: static?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  if (uart_channel_1 != NULL) {
+    _UartPolledChannel_interrupt_handler(uart_channel_1);
+  }
+}
+
+void UartPolledChannel_poll(PolledNetworkChannel *untyped_self) {
+  UartPolledChannel *self = (UartPolledChannel *)untyped_self;
+  const uint32_t minimum_message_size = 12;
```
erlingrj, line 91: Dont think this is true, should be the same as the value used above, put it in a compile def

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+        UART_CHANNEL_DEBUG("calling user callback!");
+        self->receive_callback(self->bundle, &self->output);
+      }
+    } else {
+      break;
+    }
+  }
+}
+
+unsigned int from_uc_data_bits(UartDataBits data_bits) {
```
erlingrj, line 116: static?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  case UC_UART_DATA_BITS_7:
+    return 7;
+  case UC_UART_DATA_BITS_8:
+    return 8;
+  };
+
+  return 8;
+}
+
+uart_parity_t from_uc_parity_bits(UartParityBits parity_bits) {
```
erlingrj, line 131: static?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,240 @@
...
+  case UC_UART_PARITY_MARK:
+    throw("Not supported by pico SDK");
+  case UC_UART_PARITY_SPACE:
+    throw("Not supported by pico SDK");
+  }
+
+  return UART_PARITY_EVEN;
+}
+
+unsigned int from_uc_stop_bits(UartStopBits stop_bits) {
```
erlingrj, line 148: static?

src/platform/pico/uart_channel.c
erlingrj: Ok, the UartPolledChannel pointers must be global to be accessible from the IRQs, but the FederateMessage does not need it I think

erlingrj (changes requested):
Looking mostly good! :+1: 

But we need some way of handling the initial handshake

erlingrj:
If I am not mistaken, this PR was replaced by #236 and can be closed.

PR #236, opened 2025-03-04
Bugs Found while building FSW25 Demo

tanneberger:


cmake/lfc.cmake
```diff
@@ -61,15 +61,16 @@ function(lf_build_generated_code MAIN_TARGET SOURCE_GEN_DIR)
   endif()
 
   # Check if the Include.cmake file exists in SOURCE_GEN_DIR
-  if (NOT EXISTS ${SOURCE_GEN_DIR}/Include.cmake)
+  if (NOT EXISTS ${SOURCE_GEN_DIR}/${FEDERATE}/Include.cmake)
```
erlingrj, line 64: Instead of introducing a global variable, the caller of lf_build_generated_code should pass the correct SOURCE_GEN_DIR to this function. If you code-generate a federated program, source-gen dir must include the federate name

cmake/lfc.cmake
```diff
@@ -61,15 +61,16 @@ function(lf_build_generated_code MAIN_TARGET SOURCE_GEN_DIR)
...
     message(FATAL_ERROR "Include.cmake does not exist in src-gen directory: ${SOURCE_GEN_DIR}/Include.cmake")
   endif()
 
-  include(${SOURCE_GEN_DIR}/Include.cmake)
+  include(${SOURCE_GEN_DIR}/${FEDERATE}/Include.cmake)
+  message(${REACTOR_UC_PATH})
```
erlingrj, line 69: Again, dont think we should introduce the dependency on this global variable. ALso no need to print the REACTOR_UC_PATH here

include/reactor-uc/network_channel.h
```diff
@@ -150,8 +150,11 @@ struct AsyncNetworkChannel {
 #endif
 
 #elif defined(PLATFORM_PICO)
-#ifdef NETWORK_CHANNEL_TCP_POSIX
-#error "NETWORK_POSIX_TCP not supported on PICO"
+//#ifdef NETWORK_CHANNEL_TCP_POSIX
```
erlingrj, line 153: Why is this commented out?

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
```diff
@@ -52,7 +52,8 @@ abstract class UcCmakeGenerator(
             |set(LF_MAIN_TARGET ${mainTarget})
             |set(CMAKE_BUILD_TYPE ${targetConfig.getOrDefault(BuildTypeProperty.INSTANCE)})
             |set(PLATFORM POSIX CACHE STRING "Target platform")
-            |include($S{CMAKE_CURRENT_SOURCE_DIR}/Include.cmake)
+            |#include($S{CMAKE_CURRENT_SOURCE_DIR}/src-gen/$S{LF_MAIN}/$S{FEDERATE}/Include.cmake)
```
erlingrj, line 55: We shouldnt code-generated commented out code. And this code change includes the same file but makes it more brittle by depending on a relative path

src/network_channel.c
```diff
@@ -21,9 +21,12 @@
 #endif
 
 #elif defined(PLATFORM_PICO)
-#ifdef NETWORK_CHANNEL_TCP_POSIX
-#error "NETWORK_POSIC_TCP not supported on PICO"
-#endif
+//#ifdef NETWORK_CHANNEL_TCP_POSIX
+//#error "NETWORK_POSIC_TCP not supported on PICO"
```
erlingrj, line 25: Why is this removed?

src/platform/pico/pico.c
```diff
@@ -12,7 +12,7 @@ void Platform_vprintf(const char *fmt, va_list args) { vprintf(fmt, args); }
 
 lf_ret_t PlatformPico_initialize(Platform *self) {
   PlatformPico *p = (PlatformPico *)self;
-  stdio_init_all();
+  //stdio_init_all();
```
erlingrj, line 15: Why no stdio?

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,238 @@
...
+static void UartPolledChannel_free(NetworkChannel *untyped_self) {
+  UART_CHANNEL_DEBUG("Free");
+  (void)untyped_self;
+}
+
+static bool UartPolledChannel_is_connected(NetworkChannel *untyped_self) {
+  UartPolledChannel *self = (UartPolledChannel *)untyped_self;
+  UART_CHANNEL_DEBUG("Open connection - Sending Ping");
+  char connect_message[] = UART_OPEN_MESSAGE;
+  uart_write_blocking(self->uart_device, (const uint8_t *)connect_message, sizeof(connect_message));
```
erlingrj, line 45: We shouldnt send the UART_OPEN_MESSAGE if we already are connected?

src/reaction.c
```diff
@@ -34,6 +34,7 @@ int calculate_port_level(Port *port) {
     }
   }
 
+  printf("Input port %p has level %d", port, current);
```
erlingrj, line 37: Should probably be removed

src/reaction.c
```diff
@@ -25,7 +25,7 @@ int calculate_port_level(Port *port) {
     }
   }
 
-  for (size_t i = 0; i < port->sources.size; i++) {
+  for (size_t i = 0; i < port->sources.num_registered; i++) {
```
erlingrj, line 28: num_registered and size should be the same

src/reactor.c
```diff
@@ -30,7 +30,6 @@ void Reactor_validate(Reactor *self) {
     if (trigger->type == TRIG_INPUT || trigger->type == TRIG_OUTPUT) {
       Port *port = (Port *)trigger;
       validate(port->effects.num_registered == port->effects.size);
-      validate(port->sources.num_registered == port->sources.size);
```
erlingrj, line 33: Why remove this check? It is a code-gen bug if we say that a port has N sources but then we do not register N

src/platform/pico/uart_channel.c
```diff
@@ -0,0 +1,238 @@
...
+    self->receive_buffer[self->receive_buffer_index] = received_byte;
+    self->receive_buffer_index++;
+
+    char connect_message[] = UART_OPEN_MESSAGE;
+    if (received_byte == connect_message[sizeof(connect_message) - 1] &&
+        self->receive_buffer_index >= sizeof(connect_message)) {
+      if (memcmp(connect_message, &self->receive_buffer[self->receive_buffer_index - sizeof(connect_message)],
+                 sizeof(connect_message)) == 0) {
+        self->receive_buffer_index -= sizeof(connect_message);
+        printf("Found Byte Signature of Open Message\n");
```
erlingrj, line 86: Should probably not do any printf here since we are in an ISR?

erlingrj (changes requested):
This looks mostly good. I think your solution to the handshake works.

PR #250, opened 2025-03-20
Add support for nested critical sections

erlingrj:
This is the first PR in a series of PRs to fix our concurrency model. See: https://github.com/lf-lang/reactor-uc/discussions/248

This PR adds support for nested critical sections. Since we are going to strive to keep critical sections sparse and short, it is an unnecessary complication to require us to always know whether we are in a critical section or not.


include/reactor-uc/platform.h
```diff
@@ -29,13 +29,13 @@ struct Platform {
   lf_ret_t (*wait_for)(Platform *self, interval_t duration);
   /**
    * @brief Put the system to sleep until the wakeup time or until an
-   * asynchronous event occurs.
+   * asynchronous event occurs. Must be called from a critical section.
    */
-  lf_ret_t (*wait_until_interruptible)(Platform *self, instant_t wakeup_time);
+  lf_ret_t (*wait_until_interruptible_locked)(Platform *self, instant_t wakeup_time);
```
tanneberger, line 34: does this mean before entering the function you need to lock the critical section? or does it mean this function will lock the critical section?

src/federated.c
```diff
@@ -162,6 +165,9 @@ void FederatedConnectionBundle_handle_tagged_msg(FederatedConnectionBundle *self
   // Take the value received over the network copy it into the payload_pool of
   // the input port and schedule an event for it.
   void *payload;
+
+  // FIXME: Which pool is this. WHere is `pool->free` called make sure that these are within critical sections
```
tanneberger, line 169: typo: sed/WHere/where

src/platform/riot/coap_udp_ip_channel.c
```diff
@@ -312,6 +312,7 @@ static void _client_send_blocking_callback(const gcoap_request_memo_t *memo, coa
   _lf_environment->platform->new_async_event(_lf_environment->platform);
 }
 
+// Add `_locked` or acquire critical section.
```
tanneberger, line 315: ?

tanneberger (changes requested):

include/reactor-uc/platform.h
erlingrj: `_locked` postfix means that the function only should be called when you are inside a critical section

src/platform/riot/coap_udp_ip_channel.c
erlingrj: This is a remaining TODO

src/federated.c
erlingrj: This FIXME will be solved and removed before merging

src/schedulers/dynamic/scheduler.c
```diff
@@ -179,6 +183,7 @@ static bool _Scheduler_check_and_handle_stp_violations(DynamicScheduler *self, R
     Trigger *trigger = parent->triggers[i];
     if (trigger->type == TRIG_INPUT && trigger->is_present) {
       Port *port = (Port *)trigger;
+      // FIXME: WHere is port->intended_tag updated?
```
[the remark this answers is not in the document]
erlingrj: This FIXME will be solved and removed before merging

src/platform/riot/riot.c
```diff
@@ -68,20 +68,20 @@ lf_ret_t PlatformRiot_wait_for(Platform *self, interval_t duration) {
 
 void PlatformRiot_leave_critical_section(Platform *self) {
   PlatformRiot *p = (PlatformRiot *)self;
+  p->num_nested_critical_sections--;
```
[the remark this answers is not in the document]
erlingrj: Good point

include/reactor-uc/platform.h
tanneberger: I guess having this stops making sense, when we add nested critical sections.

erlingrj: Hm, I think there still can be functions that assume that they are within a critical section rather than acquiring it self

erlingrj (commented):
This should be good-to-go now

PR #251, opened 2025-03-24
 Transient Federates

tanneberger:


include/reactor-uc/startup_coordinator.h
```diff
@@ -4,17 +4,26 @@
...
 typedef struct StartupCoordinator StartupCoordinator;
 typedef struct Environment Environment;
+typedef enum JoiningPolicy JoiningPolicy;
+typedef struct TimerConfig TimerConfig;
+
+enum JoiningPolicy { JOIN_IMMEDIATELY = 0, JOIN_ALIGNED_WITH_SHORT_TIMER = 1, JOIN_AT_HYPER_PERIOD = 2 };
 
 /** Represents the state of a neighbor. */
 typedef struct {
+  bool core_federate;                   // Whether this federate can be not available
```
edwardalee, line 19: The name of this variable and the comment contradict each other. Per the comment, if it is true, then it can be not available (is transient). Per the variable name, it's the opposite.

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcStartupCoordinatorGenerator.kt
```diff
@@ -3,9 +3,31 @@ package org.lflang.generator.uc
...
 
+enum class JoiningPolicy {
+  IMMEDIATELY,
+  TIMER_ALIGNED;
+
+  companion object {
+    fun parse(str: String): JoiningPolicy =
+        when (str) {
+          "\"IMMEDIATELY\"" -> IMMEDIATELY
+          "\"TIMER_ALIGNED\"" -> TIMER_ALIGNED
```
edwardalee, line 14: Above, these joining policies are named `JOIN_ALIGNED_WITH_SHORT_TIMER = 1, JOIN_AT_HYPER_PERIOD = 2`.
Consistency? More inconsistent names below.

src/startup_coordinator.c
```diff
@@ -55,6 +57,26 @@ static lf_ret_t StartupCoordinator_connect_to_neighbors_blocking(StartupCoordina
...
 }
 
+void StartupCoordinator_schedule_timers_joining(StartupCoordinator *self, Reactor *reactor,
+                                                interval_t federation_start_time, interval_t join_time) {
+  lf_ret_t ret;
+  for (size_t i = 0; i < reactor->triggers_size; i++) {
+    Trigger *trigger = reactor->triggers[i];
+    if (trigger->type == TRIG_TIMER) {
+      Timer *timer = (Timer *)trigger;
+      const interval_t duration = join_time - federation_start_time - timer->offset;
```
edwardalee, line 67: Should this be adding the timer->offset?

src/startup_coordinator.c
```diff
@@ -293,42 +340,174 @@ static void StartupCoordinator_handle_start_time_proposal(StartupCoordinator *se
...
+      self->msg.which_message = FederateMessage_startup_coordination_tag;
+      self->msg.message.startup_coordination.which_message = StartupCoordination_start_time_request_tag;
+      lf_ret_t ret;
+      do {
+        ret = chan->send_blocking(chan, &self->msg);
+      } while (ret != LF_OK);
+
+      // We now schedule a system event here, because otherwise we will never detect no other federates responding
+      StartupCoordinator_schedule_system_self_event(self, self->env->get_physical_time(self->env) + MSEC(250),
+                                                    StartupCoordination_start_time_response_tag);
```
edwardalee, line 367: Should MSEC(250) be a constant defined somewhere?

edwardalee (commented):
This looks very cool. I couldn't check thoroughly, but the test cases look pretty thin and I didn't see any documentation of the protocol. Minor comments and questions inserted.

src/startup_coordinator.c
tanneberger: No I dont think so here, I am calculating the time since the first firing of this time and for that I need to subtract the offset. 

I am adding the offset back two lines down when I calculate the next firing time.

include/reactor-uc/startup_coordinator.h
tanneberger: This property is relevant for how many responses to my `StartTimeRequest` I can expect from my neighbors. 

https://github.com/lf-lang/reactor-uc/pull/251/files#diff-73d218f09f4226225ac8339e6497116d217cd3ebd26cbe070501e92126de3883R418

Currently I wait until I have a response from all my neighboring core federates.

src/startup_coordinator.c
tanneberger: I now solved it by having the `TRANSIENT_WAIT_TIME` macro which can be overwritten by the user, if he wants to wait a different amount.

PR #258, opened 2025-03-28
Seperate out federated code

erlingrj:


include/reactor-uc/environments/environment_federated.h
```diff
@@ -0,0 +1,30 @@
...
+#include "reactor-uc/environment.h"
+#include "reactor-uc/network_channel.h"
+#include "reactor-uc/startup_coordinator.h"
+#include "reactor-uc/clock_synchronization.h"
+#include "reactor-uc/physical_clock.h"
+
+typedef struct EnvironmentFederated EnvironmentFederated;
+extern Environment *_lf_environment; // NOLINT
+
+struct EnvironmentFederated {
```
tanneberger, line 13: Just a tiny nitpick: I think `FederatedEnvironment` sounds nicer

include/reactor-uc/schedulers/dynamic/scheduler.h
```diff
@@ -42,6 +42,12 @@ struct DynamicScheduler {
    * tag.
    */
   void (*run_timestep)(Scheduler *self);
+
```
tanneberger, line 45: But isn't this precisely what we tried to avoid? That the stuff that we need for federated creeps into all the corners? (in this case the scheduler)

include/reactor-uc/schedulers/dynamic/scheduler_federated.h
```diff
@@ -0,0 +1,18 @@
+#ifndef SCHEDULER_FEDERATED_H
+#define SCHEDULER_FEDERATED_H
+#include "reactor-uc/queues.h"
+#include "reactor-uc/scheduler.h"
+#include "reactor-uc/schedulers/dynamic/scheduler.h"
+
+typedef struct DynamicSchedulerFederated DynamicSchedulerFederated;
+typedef struct EnvironmentFederated EnvironmentFederated;
+
+struct DynamicSchedulerFederated {
```
tanneberger, line 10: Why do we need a FederatedScheduler?

I think this will explode into our face at some point. Does this scheduler now only does the polling as an extra thing? 

if yes, I would suggest we add a new function to the Environment, like `env->bookkeeping` that based on if the environment is federated or non-federated does polling or not.

tanneberger (changes requested):

include/reactor-uc/schedulers/dynamic/scheduler_federated.h
erlingrj: Good point. We can put this function, `acquire_tag` into the Environment and provide an implementation based on whether we are federated or not

include/reactor-uc/schedulers/dynamic/scheduler.h
erlingrj: Good point!

PR #259, opened 2025-03-28
257 whether lfc should copy or symlink or not do anything the runtime should be an option

erlingrj:


tanneberger:
... the branch name

erlingrj:
Yeah it's long. Its generated by Github based on the issue. I think it is good to have branch-names prefixed by the issue-id, it makes it easier to find it later

PR #266, opened 2025-04-01
TcpIpChannel worker thread regularly wake up to check for termination

erlingrj:
I believe this PR will fix a subtle and rare race in termination. This PR adds a 1 second timeout to each call to `select`. Thus the TcpIp worker thread regularity wakes up to check whether a termination has been requested.


erlingrj:
I am going to merge this because I believe it will solve some test failures on other PRs

erlingrj:
This did not fully resolve this issue still seeing errors in CI on macOS here #268

PR #273, opened 2025-04-03
Rename lfc->lfuc

erlingrj:
To avoid confusion with the `lfc` compiler. This PR renames it to `lfuc` and `lfuc-dev`


tanneberger:
done

PR #276, opened 2025-04-03
Improve and reduce critical sections

erlingrj:
This PR is the continuation of #250 and addresses the problems discussed in #248.

I have made a pass through the code-base and I have tried reducing the critical sections to only protect the data structures that could be shared among execution contexts. I have also added the `_locked` postfix to any function that assumes that it is being called from a critical section. This makes it easer to catch code mistakes.


src/schedulers/dynamic/scheduler.c
```diff
@@ -274,27 +285,32 @@ void Scheduler_run(Scheduler *untyped_self) {
...
               start_tag);
       ArbitraryEvent e;
-      self->event_queue->pop(self->event_queue, (AbstractEvent *)&e);
+      self->event_queue->pop_locked(self->event_queue, (AbstractEvent *)&e);
       e.event.trigger->payload_pool->free(e.event.trigger->payload_pool, e.event.super.payload);
       continue;
     }
 
     if (env->poll_network_channels) {
+      env->leave_critical_section(env);
```
tanneberger, line 306: I am really not a big fan of this pattern.

what if `leave_critical_section` was called without being inside a critical section, then `enter_criticial_section` would enter even one, even if we originally were not inside a critical section

tanneberger (changes requested):
Maybe we should add to our concurrency model all the "shared" data structures and then only make those thread save. Or we should add like a custom mutex that then can protect those data-structures.

src/schedulers/dynamic/scheduler.c
erlingrj: Calling `leave_critical_section` outside a critical section is a critical error/bug. Many of the platforms will crash the program if that happens

erlingrj:
We could create a lf_mutex_t type to make the code more readable. For any bare-metal platform, the mutex would just be disable/enable interrupts. But we could still have one for each data structure and only lock the relevant mutex.

erlingrj:
But multiple mutexes also create the possibility of deadlocks.

erlingrj:
I ended up spending the entire day on this problem, because it is quite subtle. But I think this could be a better and more general solution. Here I introduce the `Mutex` type which must be implemented for each platform. The idea is to have an abstraction that can benefit from running on an OS and use an actual mutex, but also work bare-metal where all Mutex'es are implemented just by disable/enable IRQs (must take care to support nested sections though).

This PR now puts a Mutex in the following places:
1. EventQueue (the event queue has its own mutex to protect its data)
2. EventPayloadPool. ALl payload pools has a mutex to make allocate/free thread-safe. For some events, this was already the case, but for physical actions, FederatedInputPorts, SystemEvents it was not the case.
3. FederatedInputConnection. Each of them get an Mutex to protect the `last_known_tag` fields.
4. Scheduler. The scheduler needs a mutex because `current_tag`, `stop_tag`, `start_time` are read from async and channel context and modified from runtime context.

A possible optimization would be to make MUTEX_UNLOCK and LOCK intoa NOOP if we have only a single execution context. We could that by routing lock/unlock through the environment.

include/reactor-uc/clock_synchronization.h
```diff
@@ -55,7 +56,7 @@ struct ClockSynchronization {
   interval_t period;              // The period between sync request messages are sent to the neighbor master.
   ClockSyncTimestamps timestamps; // The timestamps used to compute clock offset.
   ClockServo servo;               // The PID controller
-  // TODO: Is this message here needed? Why cant we use the one on the bundle?
+  // FIXME: Is this message here needed? Why cant we use the one on the bundle?
```
tanneberger, line 59: Depends. Currently, we don't have a clear policy where the FederatedMessage struct should life. 

At some places you use the FederatedMessage struct from the Channel, sometimes you have your own.

include/reactor-uc/environment.h
```diff
@@ -19,9 +19,10 @@ typedef struct Environment Environment;
 extern Environment *_lf_environment; // NOLINT
 
 struct Environment {
-  Reactor *main;         // The top-level reactor of the program.
-  Scheduler *scheduler;  // The scheduler in charge of executing the reactions.
-  Platform *platform;    // The platform that provides the physical time and sleep functions.
+  Reactor *main;        // The top-level reactor of the program.
+  Scheduler *scheduler; // The scheduler in charge of executing the reactions.
+  Platform *platform;
+  PLATFORM_T _platform;  // The platform that provides the physical time and sleep functions.
```
tanneberger, line 25: would make code-generation easier ...

src/platform/flexpret/flexpret.c
```diff
@@ -56,60 +45,59 @@ lf_ret_t PlatformFlexpret_wait_for(Platform *super, interval_t wait_time) {
...
+void Platform_ctor(Platform *super) {
+  PlatformFlexpret *self = (PlatformFlexpret *)super;
+  super->get_physical_time = PlatformFlexpret_get_physical_time;
+  super->wait_until = PlatformFlexpret_wait_until;
+  super->wait_for = PlatformFlexpret_wait_for;
+  super->wait_until_interruptible = PlatformFlexpret_wait_until_interruptible;
+  super->notify = PlatformFlexpret_notify;
+  self->num_nested_critical_sections = 0;
+  self->async_event_occurred = false;
+  Mutex_ctor(&self->mutex.super);
```
tanneberger, line 62: Why is there a mutx inside the platform struct?

Issue #277, opened 2025-04-03
Implement lf_sleep


edwardalee:
There appears to be no implementation of `lf_sleep` in reactor-uc.

erlingrj:
We could expose the `platform->wait_for` function which waits/sleeps for a duration. It does not return early if there is an async event such as `wait_until`

erlingrj:
Should `env->wait_for(env)` (AKA `lf_sleep`) be affected by `fast: true`?

PR #278, opened 2025-04-03
Add env->wait_for

erlingrj:
Closing #277

### DIFF
--- a/src/environments/unfederated_environment.c
+++ b/src/environments/unfederated_environment.c
@@ -37,6 +37,14 @@ static lf_ret_t Environment_wait_until_locked(Environment *self, instant_t wakeu
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/environments/unfederated_environment.c
```diff
@@ -37,6 +37,14 @@ static lf_ret_t Environment_wait_until_locked(Environment *self, instant_t wakeu
   }
 }
 
+static lf_ret_t Environment_wait_for(Environment *self, interval_t wait_time) {
```
tanneberger, line 40: Maybe some docs string would be nice, because this is technically user-facing api.

tanneberger (changes requested):

src/environments/unfederated_environment.c
erlingrj: The doc strings are in the header file

PR #285, opened 2025-04-23
[WIP] Enclaved reactors

erlingrj:
This PR introduces enclaves into reactor-uc. To do so we must extend the platform API with a notion of threads, because each enclave executes in its own execution context. Currently the threading API is only implemented in POSIX and enclaves can thus still be considered experimental. 

Enclaves behave as federates and are coordinated decentrally. This means that we could get STP violations and that we do not support ZDC cycles between enclaves


erlingrj:
I realized there is a couple of things missing:
- [ ] Support communication between a deeply nested enclave and the "main" environment higher up. Currently only enclave -> enclave support
- [ ] Figure out keep_alive, has_async_events for federated, enclaved and normal. We might need to rename these. 
- [ ] Throw error if we try to register a reaction triggered directly by an enclave port.
- [ ] Throw error if we put a trigger or a reaction in enclaved or federated reactor.
- [ ] Figure out how to properly do shutdown by starvation. This can be done when all inputs coming from other enclaves have last_known_tag of FOREVER

PR #287, opened 2025-05-13
target_include_directories for src Folder

tanneberger:
so you can actually include header files from your src folder.

### DIFF
--- a/lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
+++ b/lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
@@ -50,6 +50,7 @@ abstract class UcCmakeGenerator(
             |cmake_minimum_required(VERSION $minCmakeVersion)
             |project(${mainTarget} LANGUAGES C)
             |set(LF_MAIN_TARGET ${mainTarget})
+            |set(SOURCE_FOLDER $S{CMAKE_SOURCE_DIR}/../../../src)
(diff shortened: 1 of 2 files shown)


lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
```diff
@@ -50,6 +50,7 @@ abstract class UcCmakeGenerator(
             |cmake_minimum_required(VERSION $minCmakeVersion)
             |project(${mainTarget} LANGUAGES C)
             |set(LF_MAIN_TARGET ${mainTarget})
+            |set(SOURCE_FOLDER $S{CMAKE_SOURCE_DIR}/../../../src)
```
erlingrj, line 53: This makes some assumptions about the project layout. I think instead we should leverage that we have the path to the LF source file available at compile time. I think the include path should be computed based on that path instead

erlingrj (changes requested):
That makes sense. But can you do it without relying on a relative path which might not be correct?

tanneberger:
@erlingrj this is just the default value, the path will be always set when lfc invokes this cmake file

erlingrj:
Sorry for the delay.

> @erlingrj this is just the default value, the path will be always set when lfc invokes this cmake file.

I think it would be better that lfc puts the value directly into the CMakeLists, and not just passing it as an argument to cmake. Why not put the actual value as the default, and not that relative path?

PR #290, opened 2025-06-25
add break in StartupCoordinationState_RUNNING case to stop implicit fallthrough

SheaFixstars:
Added a break into the `startup_coordinator.c`  `StartupCoordinator_handle_start_time_proposal` switch case. Without this the Zephyr build is broken due to a  `-Werror=implicit-fallthrough=` error. 

The case will exit the program if hit so the added break does not affect logical operations at all other than removing the warning/error.

```
In file included from /home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/include/reactor-uc/startup_coordinator.h:4,
                 from /home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/src/startup_coordinator.c:1:
/home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/src/startup_coordinator.c: In function ‘StartupCoordinator_handle_start_time_proposal’:
/home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/include/reactor-uc/error.h:32:8: error: this statement may fall through [-Werror=implicit-fallthrough=]
   32 |     if ((expr) == 0) {                                                                                                 \
      |        ^
/home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/src/startup_coordinator.c:253:7: note: in expansion of macro ‘validate’
  253 |       validate(false);
      |       ^~~~~~~~
/home/shea/git/lf-zephyr-uc-template/src-gen/HelloWorld/reactor-uc/src/startup_coordinator.c:254:5: note: here
  254 |     case StartupCoordinationState_HANDSHAKING:
      |     ^~~~
cc1: all warnings being treated as errors
ninja: build stopped: subcommand failed.
FATAL ERROR: command exited with status 1: /usr/bin/cmake --build /home/shea/git/lf-zephyr-uc-template/build --target run
```

### DIFF
--- a/src/startup_coordinator.c
+++ b/src/startup_coordinator.c
@@ -251,6 +251,7 @@ static void StartupCoordinator_handle_start_time_proposal(StartupCoordinator *se
     case StartupCoordinationState_RUNNING:
       // Should not be possible.
       validate(false);
+      break;
     case StartupCoordinationState_HANDSHAKING:
       // This is possible. Our node might be still handshaking with another neighbor.
       // Intentional fall-through


erlingrj:
This looks good. There is an alarming test failure in the CI on macos in one of the federated programs. It suggests that there is a bug with the startup coordinator. This is the assertion failure that trips the program:

```
Assertion failed at  /Users/runner/work/reactor-uc/reactor-uc/test/lf/src-gen/FederatedChain/r2/reactor-uc/src/startup_coordinator.c:261
```

This means that it is a mismatch between the number of start time proposals that the federate has recorded and the number of start time proposal that the sending federate claims to have sent. This will require more debugging to get to the bottom of. I will try to reproduce locally


The other failures are because the CI action tries to comment on the PR and it does not have the right priveliges, probably because it is coming from a fork?

erlingrj (approved):

edwardalee:
What is the status of this PR? It seems necessary for Zephyr to work, but there are odd test failures.  According to Copilot, the test errors are because of this:

The job failed due to a permissions error when trying to create an issue comment:
Resource not accessible by integration

This error typically happens when a GitHub Actions workflow tries to interact with the repository using the GitHub Actions token, but the required permissions are missing. In this case, the job attempted to create an issue comment through the GitHub REST API and was denied.

Solution:

Update your workflow file (.github/workflows/coverage.yml) and explicitly set permissions for the job or workflow to allow the necessary access. For posting comments, you need write access to issues and pull requests.
Example change for your workflow file:

YAML
```
permissions:
  issues: write
  pull-requests: write
```
Place this at the top level of the workflow YAML file, ideally just after name:.

If you’re using a custom token, ensure it has repository write permissions.

If using an action to post comments, check that it’s being invoked with the right token (usually ${{ secrets.GITHUB_TOKEN }}) and that the token is not being overridden or restricted elsewhere.

You do not need to change your tests or code; the actual tests passed (100% tests passed, 0 tests failed out of 16). The failure is in the workflow configuration, not your source code or tests.

For more information, see the GitHub documentation:
https://docs.github.com/rest/issues/comments#create-an-issue-comment

Update your workflow and re-run the job. This should resolve the failing post-job step.

PR #297, opened 2025-09-13
Redo how maxwait is specified and used

edwardalee:
This PR changes how `maxwait` is specified. It includes:
* Specifying as an attribute on an instantiation, as in:
   ```
     @maxwait(forever)
     a = new Foo()
   ```
   This is an STA.
* Specifying maxwait on a connection provides _additional_ wait time for the destination port(s) over what is specified on the instantiation.
  ```
  @maxwait(100 ms)
  b.out -> a.in
  ```
  This is an STAA.
* Updating the validator to check the syntax of the attribute
* Requiring `maxwait` on reactions to have no numerical value, e.g.
   ```
     reaction(in) {= ... =} maxwait {= ... =}
   ```
* Major refactoring of the AttributeUtils class, including removing unused methods.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,7 +14,8 @@ Checks: '
     -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,
     -readability-function-cognitive-complexity,
     -modernize-macro-to-enum,
-    -readability-math-missing-parentheses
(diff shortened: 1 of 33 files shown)


tanneberger (approved):
Maybe add this small kotlin change and then I think this is ready to be merged.

PR #306, opened 2025-09-24
FederatedEnvironment: Runtime didn't respect largest max-wait

tanneberger:
If there there multiple NetworkBundles, only the last and not largest max-wait value was used.


edwardalee:
This looks reasonable, but `MAX` needs to be defined, and I think this will lead to test failures because of [this discussion on Zulip](https://lf-lang.zulipchat.com/#narrow/channel/460540-reactor-uc/topic//with/541290722).

PR #307, opened 2025-09-29
Remove multiThreaded Attribute

tanneberger:


edwardalee (approved):
LGTM.  This also brings this file into alignment with lingua-franca/master.

PR #319, opened 2025-11-27
lf_schedule now has a return value

tanneberger:
1.) Will return `LF_OUT_OF_BOUNDS` when the event queue is full
2.) Will return `LF_NO_MEM` if the value buffer is full

### DIFF
--- a/include/reactor-uc/macros_api.h
+++ b/include/reactor-uc/macros_api.h
@@ -54,26 +54,21 @@
  */
 #define lf_is_present(trigger) (((Trigger*)(trigger))->is_present)
 
+lf_ret_t lf_schedule_with_value(Action* action, interval_t offset, const void* val);
+
 /// @private
 #define lf_schedule_with_val(action, offset, val)                                                                      \
-  do {                                                                                                                 \
+  ({                                                                                                                   \
     __typeof__(val) __val = (val);                                                                                     \
-    Action* __a = (Action*)(action);                                                                                   \
-    lf_ret_t ret = __a->schedule(__a, (offset), (const void*)&__val);                                                  \
-    if (ret == LF_FATAL) {                                                                                             \
-      LF_ERR(TRIG, "Scheduling an value, that doesn't have value!");                                                   \
-      Scheduler* sched = __a->super.parent->env->scheduler;                                                            \
-      sched->do_shutdown(sched, sched->current_tag(sched));                                                            \
-      throw("Tried to schedule a value onto an action without a type!");                                               \
-    }                                                                                                                  \
-  } while (0)
+    lf_schedule_with_value((Action*)action, offset, &__val);                                                           \
+  })
 
 /// @private
 #define lf_schedule_without_val(action, offset)                                                                        \
-  do {                                                                                                                 \
+  ({                                                                                                                   \
     Action* __a = (Action*)(action);                                                                                   \
     __a->schedule(__a, (offset), NULL);                                                                                \
-  } while (0)
+  })
 
 /// @private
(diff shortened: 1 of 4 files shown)


include/reactor-uc/macros_api.h
```diff
@@ -54,26 +54,21 @@
...
-    Action* __a = (Action*)(action);                                                                                   \
-    lf_ret_t ret = __a->schedule(__a, (offset), (const void*)&__val);                                                  \
-    if (ret == LF_FATAL) {                                                                                             \
-      LF_ERR(TRIG, "Scheduling an value, that doesn't have value!");                                                   \
-      Scheduler* sched = __a->super.parent->env->scheduler;                                                            \
-      sched->do_shutdown(sched, sched->current_tag(sched));                                                            \
-      throw("Tried to schedule a value onto an action without a type!");                                               \
-    }                                                                                                                  \
-  } while (0)
+    lf_schedule_with_value((Action*)action, offset, &__val);                                                           \
```
edwardalee: It's not clear to me why `__typeof__(val) __val = (val)` is needed since `lf_schedule_with_value` takes a `void*`. In fact, why is `lf_schedule_with_val` needed?  It seems kind of risky, in fact, because it hides the fact that a pointer is passed to `lf_schedule_with_value`. Users are likely to use `lf_schedule_with_val` with the value argument being on the stack, which will not work.

src/macro_api.c
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @brief this is a helper function for the lf_schedule macro with a value
+ **/
+lf_ret_t lf_schedule_with_value(Action* action, interval_t offset, const void* val) {
+  lf_ret_t ret = action->schedule(action, offset, val);
+  if (ret == LF_FATAL) {
+    LF_ERR(TRIG, "Scheduling a value, for an void action!");
+    Scheduler* sched = action->super.parent->env->scheduler;
+    sched->do_shutdown(sched, sched->current_tag(sched));
+    throw("Tried to schedule a value onto an action without a type!");
```
edwardalee: Are two error messages needed here?  I'm not sure what `throw` does.

edwardalee (approved):
Looks like an improvement, but I have some questions.

src/macro_api.c
tanneberger: `throw` on many embedded platforms halts the execution and prints this message. I think this is what we want, because scheduling a void-action with a value, should be a hard crash.

test/lf/src/ActionOverflow.lf
```diff
@@ -0,0 +1,23 @@
...
+      if (lf_schedule(a, SEC(1), 0) != LF_OK) {
+        self->discarded_schedules++;
+      }
+  =}
+
+  reaction(a) {=
+    if (self->discarded_schedules == 90) {
+      printf("Action is now executed, discarded %i events in the process!\n", self->discarded_schedules);
+      env->request_shutdown(env);
+    }
```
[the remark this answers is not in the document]
tanneberger: Agreed, fixed and pushed.

include/reactor-uc/macros_api.h
tanneberger: small additional note the `lf_schedule_with_val` and `lf_schedule_without_val` are for internal use only (see the `@private` in-front) they are used by the `lf_schedule` macro, which is variadic  (see the `LF_SCHEDULE_CHOOSER` magic)

src/macro_api.c
edwardalee: I cannot find any reference to `throw` in any C standard. Can you provide a pointer?

PR #328, opened 2026-01-02
[DRAFT] Priority-based scheduler for federated applications

fra-p:
This draft pull request is the work @edwardalee and I have been doing to introduce priority scheduling of reactions with deadline for federated applications with the reactor-uc runtime. It is still a draft, because testing is incomplete and the syntax for some customization options is to be discussed (more on this in a dedicated section). However, I am creating this pull request to get feedback on the priority assignment rule and the scheduler design that would certainly help improve the scheduler. Feel free to comment with thoughts and suggestions!

# Introduction
The reactor-uc runtime features a single OS thread that schedules the reactions in the reaction queue. In federated applications, each federate has its own runtime and its own thread. When the federates run on the same machine, the threads compete for the CPUs of the platform, and a scheduling algorithm handles contention.

Federates may have different urgency of execution depending on what reaction they are executing at a given time. In fact, reactions can be assigned with a deadline. To guarantee that deadlines are met, a real-time thread scheduler is needed to prioritize the completion of reactions with more urgent deadlines.

This PR implements a priority-based scheduler in reactor-uc that assigns a priority value to a reaction as a function of its deadline. The rule is based on the deadline monotonic (DM) principle, but extended to support platforms like Linux that feature a finite number of assignable priority values.

# Scheduler design

## Control flow
The main thread of each federate serves one reaction at a time. The thread picks the first ready reaction in the reaction queue and sets its priority to a value that depends on the deadline of the picked reaction. The OS thread scheduler will use that priority value to schedule the thread together with the threads of the other federates on the same machine.

Once a federate is done with the reactions in the reaction queue and waits until its physical time catches up with the next tag, it sets its priority to a value that is higher than any possible main thread executing reactions. This is needed to handle the case of asynchronous events (e.g., physical connections) and guarantees that the thread is woken up as soon as the event arrives.

Federates rely on TCP connections to communicate, and the connections are handled by the TCP thread. Similarly to the previous case, the priority of the TCP thread is higher than any sleeping thread (and thus, any thread executing reactions), to be sure that the messages are received without delays. However, this might be a problem with large messages.

## Priority function
The priority value for a thread executing a reaction with deadline $d$ is determined by the priority function $prio(d)$, which maps the deadline range [$d_{min}, d_{max}$] to the priority range [${prio}_{min}, {prio}_{max}$]. $d_{min}$ and $d_{max}$ are the minimum and maximum deadlines found in a given federated application, while ${prio}_{min}$ and ${prio}_{max}$ are the minimum and maximum scheduling priorities that can be assigned to threads scheduled with a given OS thread scheduler. The function is decreasing exponential to subsume the deadline monotonic rule: tighter deadlines are mapped to higher priority values. This function is tailored to the specific federated application and generated by the code generator.

The analytical expression of the priority function is:

$prio(d) = round(K e^{- \alpha * d} + P)$.

$\alpha$ is the parameter that controls the speed of decay of the exponential function. $K$ and $P$ are computed such that:
- $prio(d_{min}) = {prio}_{max}$;
- $prio(d_{max}) = {prio}_{min}$.

Since the priority space is finite, collisions may occur: two very close deadline values may be mapped to the same priority value. The priority function allows to control in what part of the deadline range collisions are more likely to occur by tuning the $\alpha$ parameter. A big $\alpha$ value increases the speed of decay of the function, which means that most of the priority space is used to map small deadlines. This is useful when most of the deadlines of the federated application are tight and we do not want collisions to occur with these. Collisions occurring with looser deadlines are likely to cause less harm. A small $\alpha$ value decreases the speed of decay, and the priority function performs a priority mapping that tends to a linear conversion. $\alpha$ is automatically selected in the range [$\alpha_{min}$, $\alpha_{max}$] (empirically derived) by the code generator as a function of the median deadline of the federated application, to make the priority function better fit its deadline distribution.

## Implementation details
The priority scheduler is a sublcass of the dynamic scheduler and is called `PriorityScheduler`. It only works with platforms having threading capabilities that are modeled by the `ThreadedPlatform` class.

The only supported platform at the moment is Linux. Both `SCHED_FIFO` and `SCHED_RR` can be selected as underlying thread scheduling algorithms. The POSIX thread API is used to control the thread scheduler.

Federate's main threads can be assigned with priority values in the range [2, 97]. 1 is reserved to reactions with no deadline; 98 is the sleeping thread's priority; 99 is the TCP thread's priority.

The `PriorityScheduler_run_timestep` function sets the priority of the current thread right before executing the reaction body by calling the priority function.

# Example
[Tasks.lf.txt](https://github.com/user-attachments/files/24405943/Tasks.lf.txt)
Attached is a simple example of a federated real-time application in which deadline violations occur when it is not executed with the priority scheduler. It is made of two periodic tasks both triggered by a 50 ms timer, one of which has a tighter deadline and a longer execution time than the other. The more urgent task starts at a slightly greater logical time, which means that when its timer fires, the loose-deadline task is already executing. This test case shows the ability of the priority scheduler to preempt the loose-deadline task to execute the more urgent one. With the default scheduler, preemption does not happen and the urgent task misses its deadline.

# Future work
Currently, the priority scheduler is enabled when either `SCHED_FIFO` or `SCHED_RR` are selected as Linux scheduling policy. The selection is done with the *thread-policy* target property that takes as values "rt-fifo", "rt-rr" and "normal" (for the default non-real-time Linux scheduler). Similarly, the *cores* target property restricts the number of cores that the application can use. In the future, these two target properties will become configuration properties specified in the main reactor with the '@' sign. The most appropriate syntax is yet to be defined.

A more thorough testing is in the works to make sure all corner cases are handled correctly.


lfc/core/src/main/java/org/lflang/target/property/CoresProperty.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @Override
+  public Integer initialValue() {
+    return 0;
+  }
+
+  @Override
+  protected Integer fromString(String string, MessageReporter reporter) {
+    return Integer.parseInt(string); // FIXME: check for exception
```
edwardalee: Deal with FIXME here?

lfc/core/src/main/java/org/lflang/target/property/type/ThreadPolicyType.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Enumeration of thread policies to be used to parameterize the thread scheduler.
+   *
+   * @author Shaokai Lin
+   * @author Marten Lohstroh
+   */
+  public enum ThreadPolicy {
+    NORMAL("normal", "LF_SCHED_FAIR"),
+    REALTIME_RR("rt-rr", "LF_SCHED_TIMESLICE"),
+    REALTIME_FIFO("rt-fifo", "LF_SCHED_PRIORITY");
```
edwardalee: Why not make these match the C names?

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcMainGenerator.kt
```diff
@@ -26,17 +43,233 @@ abstract class UcMainGenerator(
 
   abstract fun keepAlive(): Boolean
 
+  /**
+   * Collects all deadlines from federates, including nested reactors.
```
edwardalee:
```suggestion
   * Collect all deadlines from federates, including nested reactors.
```

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcMainGenerator.kt
```diff
@@ -26,17 +43,233 @@ abstract class UcMainGenerator(
...
+   * This method traverses through all federates and their nested reactor instantiations
+   * to find all reactions with deadlines.
+   * 
+   * @param federates List of federates to collect deadlines from
+   * @return List of all deadlines found in the federates and their nested reactors
```
edwardalee: Would it make more sense for the return value to be a Set rather than a List? A list has semantics to the ordering and has duplicate entries.  Same for all related methods below.

edwardalee (changes requested):
I've only been through part of this PR, but I found some issues, so I'm submitting a partial review in case these comments are useful. Most of the comments are insignificant... minor suggestions for documentation. But one is potentially a bug.

tanneberger (commented):
Firstly: Very Good Work! :tada: :+1: :wrench: 

I left a lot of comments (sorry :D)

Couple of more big picture comments (it's okay if you hate me XD):

- I would like to remove target properties from uC => turn them into annotations, this would allow us to have federates running on embedded (zephyr/riot) talking to a LinuxRT federate. 

- The code generator is right now is the ugliest part of uC: Put the Priority stuff into a `PriortyScheduler.kt` and lets add a good integration

- I would recommend we add a `PosixRT` platform that implements the `ThreadedPlatform`, we still want that @edwardalee can run uC on his Macbook where there is no LinuxRT scheduler :D

PR #334, opened 2026-02-10
Add action scheduling policies

### LINKED ISSUE #173: Implement "drop" and "replace" policy for actions
#170 added support for min spacing of events scheduled on actions. It uses the default "defer" policy to handle violations. LF also defines "drop" and "replace" which we should implement. "drop" is very simple, simply  drop the event and return. "replace" requires us to search through the event queue to find the last (?) pending event of an action and the replace the payload of this event (also we must remember to free the old payload). This will require some addition to the EventQueue API, but nothing too bad.

sbgaia:
This PR implements action scheduling policies (`drop`, `defer`, `replace`, and `update`) that control how events are handled when they violate the min_spacing constraint on an action.

Major Changes:
- Implement policy-based action scheduling logic
- Replace recursive heapify with an iterative implementation in the event queue
- Add new event queue tests to verify that the new methods are working properly

Failing Tests:
The `action_overwrite_test` is now failing. The previous behavior of the defer policy was not compliant with reactor-c. 
In reactor-c, when a `defer` policy is used with `min_spacing` set to 0, duplicate events are deferred to the next microstep rather than replacing the current value. This behavior was introduced here (#111). Is there a specific reason for the change in semantics?

Closes #173


src/action.c
```diff
@@ -64,15 +65,55 @@ lf_ret_t Action_schedule(Action* self, interval_t offset, const void* value) {
...
+
+    case update:
+      if (found != NULL) {
+        LF_DEBUG(TRIG, "Updating event on action %p. Removing old event scheduled at time (%lld, %d).", self,
+                 found->super.tag.time, found->super.tag.microstep);
+        lf_ret_t ret = event_queue->remove(event_queue, &found->super);
+        event_queue->remove(event_queue, &found->super);
+        validate(ret == LF_OK);
+      }
+      break;
```
edwardalee: This seems to result in going ahead and scheduling the event if `found` is NULL, but this could violate the `min_spacing` policy. Is this what we want? Tagging @lsk567 since he created this policy.

src/queues.c
```diff
@@ -46,35 +46,101 @@ static lf_ret_t EventQueue_insert(EventQueue* self, AbstractEvent* event) {
 
   memcpy(&self->array[self->size], event, event_size);
 
-  if (self->size++ > 0) {
-    for (int i = ((int)self->size) / 2 - 1; i >= 0; i--) {
-      self->heapify_locked(self, i);
+  self->size++;
+  size_t idx = self->size - 1;
```
edwardalee:
```suggestion
  size_t idx = self->size++;
```

edwardalee (changes requested):
This is great!  I'm not sure the "update" policy is quite right, though. It seems it can result in violating the `min_spacing`.  Tagging @lsk567 on this since the documentation at lf-lang.org is missing for this policy.

edwardalee:
You have identified a legitimate bug in reactor-c. Can you open an issue there?

sbgaia:
> You have identified a legitimate bug in reactor-c. Can you open an issue there?

Done here: https://github.com/lf-lang/reactor-c/issues/560

tanneberger (changes requested):

include/reactor-uc/action.h
```diff
@@ -12,9 +12,25 @@ typedef struct LogicalAction LogicalAction;
...
+ * time is satisfied.
+ * The `drop` policy simply drops events that are scheduled too early.
+ * The `replace` policy will attempt to replace the payload of the preceding event.
+ * Unless the preceding event has already been handled, it gets assigned the value
+ * of the new event. If the preceding event has already been popped off the event
+ * queue, the `defer` policy is fallen back to.
+ * The `update` policy drops the preceding event, if it is still in the event queue,
+ * and updates it with the newly scheduled event.
+ */
+typedef enum { defer, drop, replace, update } ActionPolicy;
```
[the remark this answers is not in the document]
tanneberger: and maybe consider adding a prefix like `ACTION_POLICY_DEFER`

include/reactor-uc/scheduler.h
```diff
@@ -57,6 +57,18 @@ struct Scheduler {
...
+   * @brief Cancel a pending event for a `trigger` at `event_time`. Returns
+   * LF_NOT_FOUND if no matching event exists.
+   */
+  lf_ret_t (*cancel_event)(Scheduler* self, Trigger* trigger, instant_t event_time);
+
+  /**
+   * @brief Replace the payload of a pending event for a `trigger` at
+   * `event_time`. Returns LF_NOT_FOUND if no matching event exists.
+   */
+  lf_ret_t (*replace_event_payload)(Scheduler* self, Trigger* trigger, instant_t event_time, const void* new_value);
```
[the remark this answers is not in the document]
sbgaia: Yes, we can discuss this. We already have this method in the event queue:

```c
ArbitraryEvent* (*find_equal_same_tag)(EventQueue* self, AbstractEvent* event);
```
We could add this method to the scheduler interface. Also, I was thinking we could update the `lf_schedule` macro to return either the created `Event` or its `tag_t`. This way, users would be able to cancel queued events directly by calling `sched->cancel_event(...)`.

src/action.c
tanneberger: here I would also expect that the defer policy is applied.

tanneberger (commented):
Very good work!

Looks very close to being ready to merge :)

the only more major things I was confused that it appears you only trigger the defer policy only when the min-spacing is violated.

tanneberger (approved):

PR #335, opened 2026-02-13
self-hosted

EhsanKhodadad:
self-hosted Patmos runner is added


src/platform/patmos/patmos.c
```diff
@@ -122,10 +122,15 @@ Platform* Platform_new(void) { return (Platform*)&platform; }
 void MutexPatmos_unlock(Mutex* super) {
   (void)super;
   PlatformPatmos* platform = (PlatformPatmos*)_lf_environment->platform;
+  if (platform->num_nested_critical_sections == 0) {
+    LF_ERR(PLATFORM, "MutexPatmos_unlock underflow before decrement");
```
tanneberger: maybe also throw

PR #336, opened 2026-02-15
Optimize generation of Multiport and Bank connection statements.

tanneberger:


lfc/core/src/main/kotlin/org/lflang/generator/uc/UcConnectionGenerator.kt
```diff
@@ -370,8 +371,80 @@ class UcConnectionGenerator(
...
+      groupedConn: UcGroupedConnection,
+      src: UcChannel,
+      dest: UcChannel,
+      size: Int,
+  ) =
+      """|lf_grouped_connect((Connection *) &self->${groupedConn.getUniqueName()}[${src.getCodeBankIdx()}][${src.getCodePortIdx()}], (Port *) ${src.generateChannelPointer()}, (Port *) ${dest.generateChannelPort()}, ${size});
+    """
+          .trimMargin()
+
+  private fun generateInterleavesConnectChannel(
```
edwardalee: Should this be `generateInterleavedConnectChannel`?

edwardalee (changes requested):
It's hard for me to review this without comments. I can't tell what the functions are intended to do, so it's hard to evaluate their implementation. For example, what is the difference between `generateConnectionStatements` and `generateOutboundConnection`? What is an "outbound connection" vs. a "connection statement"?  It's also not clear to me what `lf_grouped_connect` is supposed to do. It looks like it is meant to support a subset of LF group connections?  I.e., I'm guessing it supports:
```
  foo.out -> bar.in, baz.in
```
but not
```
  foo.out, faz.out -> bar.in, baz.in
```

PR #342, opened 2026-03-19
Coordinated Shutdown

tanneberger:
This pull request introduces coordinated shutdown logic to the federation protocol. Similar to the startup synchronization process, this ensures that all federates terminate at a consistent, globally agreed-upon logical time.

## Key Changes

- Updated API: The request_shutdown interface now supports a time offset. Users can specify when a shutdown should occur relative to the current logical time.  e.g. `env->request_shutdown(env, SEC(1))` requests a shutdown one second into the future.
- Shutdown Broadcasting: When a federate requests a shutdown, the proposed time is broadcasted across the federation.
- Consensus Logic: To resolve conflicts between multiple shutdown requests: The earliest announcement containing the latest shutdown time wins the broadcast and is established as the stop_tag. This logic mirrors the existing startup protocol.

## Important Notes & Caveats

- Once a stop_tag has been established and broadcasted, it cannot be overwritten or delayed by subsequent requests.
- The stop_tag is inclusive. All events scheduled at the exact logical time of the stop_tag will be processed before the reactor shuts down.


erlingrj:
Very cool Tassilo! I will try to review this soon. How do you handle the period between the time t1 when a shutdown time has been proposed (either by the federated or when receiving a message with shutdown request from a neighbor) and the time when it is confirmed (after the designated number of steps of the gossip protocol).

Do you stall the advancement of logical time? This could create a deadline misses unless our max-lags also apply gere.

include/reactor-uc/macros_internal.h
```diff
@@ -572,6 +572,24 @@ typedef struct FederatedInputConnection FederatedInputConnection;
...
+#define LF_DEFINE_SHUTDOWN_COORDINATOR_STRUCT(ReactorName, NumNeighbors, NumEvents)                                    \
+  typedef struct {                                                                                                     \
+    ShutdownCoordinator super;                                                                                         \
+    ShutdownEvent events[(NumEvents)];                                                                                 \
+    bool used[(NumEvents)];                                                                                            \
+  } ReactorName##ShutdownCoordinator;
+
+#define LF_DEFINE_SHUTDOWN_COORDINATOR_CTOR(ReactorName, NumNeighbors, LongestPath, NumEvents)                         \
+  void ReactorName##ShutdownCoordinator_ctor(ReactorName##ShutdownCoordinator* self, Environment* env) {               \
+    ShutdownCoordinator_ctor(&self->super, env, NumNeighbors, sizeof(ShutdownEvent), (void*)self->events, self->used,  \
```
sbgaia: The third argument of the ctor shouldn't be LongestPath?

src/shutdown_coordinator.c
```diff
@@ -0,0 +1,119 @@
...
+  ShutdownCoordinator* self = (ShutdownCoordinator*)_self;
+  ShutdownEvent* payload = (ShutdownEvent*)event->super.payload;
+  switch (payload->msg.which_message) {
+  case ShutdownCoordination_shutdown_time_announcement_tag:
+    LF_INFO(FED, "Handle: ShutdownCoordination time proposal tag");
+    ShutdownCoordinator_handle_time_proposal(self, &payload->msg, payload->neighbor_index);
+    break;
+  default:
+    throw("Unknown shutdown message, which cannot be handled!");
+  }
```
sbgaia, line 101: Shouldn't we free the payload here (the logic seems similar to the StartupCoordinator)?
```suggestion
  }
  _self->payload_pool.free(&_self->payload_pool, event->super.payload);
```

src/shutdown_coordinator.c
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+
+    self->env->scheduler->request_shutdown(self->env->scheduler, self->proposed_shutdown_time);
+  } else {
+    LF_WARN(FED, "Shutdown message received but dropped! Current Step: %i", current_step);
+  }
+}
+
+/** Invoked by scheduler when handling any system event destined for StartupCoordinator. */
```
sbgaia:
```suggestion
/** Invoked by scheduler when handling any system event destined for ShutdownCoordinator. */
```

src/shutdown_coordinator.c
```diff
@@ -0,0 +1,119 @@
...
+                                  .microstep = msg->message.shutdown_time_announcement.current_tag.microstep};
+  const tag_t shutdown_tag = {.time = msg->message.shutdown_time_announcement.shutdown_tag.time,
+                              .microstep = msg->message.shutdown_time_announcement.shutdown_tag.microstep};
+
+  size_t current_step = msg->message.shutdown_time_announcement.step;
+  bool shutdown_larger = lf_tag_compare(shutdown_tag, self->proposed_shutdown_time) == 1;
+  bool announcement_time_equal = (bool)(lf_tag_compare(announcement_tag, self->announcement_of_shutdown) == 0 ||
+                                        lf_tag_compare(self->announcement_of_shutdown, NEVER_TAG) == 0);
+
+  if (shutdown_larger && announcement_time_equal && current_step <= self->longest_path) {
```
sbgaia: I have a doubt about this condition. If two reactors propose a shutdown at the same logical time but with different microsteps, this could drop messages and cause federates to stop at different logical times.

For example:
r1 -> at (t,0) proposes shutdown (t+2s, 0)
r2 -> at (t,1) proposes shutdown (t+3s, 0)

Since this condition checks for strict equality between tags, the check will fail. This means the final stop tag becomes non-deterministic and entirely dependent on the order in which these messages are received. Is this intended, or should we just check the time (as the variable name suggests)?

src/schedulers/dynamic/scheduler.c
```diff
@@ -456,14 +462,14 @@ lf_ret_t Scheduler_schedule_system_event_at(Scheduler* super, SystemEvent* event
...
-void Scheduler_request_shutdown(Scheduler* untyped_self) {
+void Scheduler_request_shutdown(Scheduler* untyped_self, tag_t shutdown_time) {
   DynamicScheduler* self = (DynamicScheduler*)untyped_self;
 
   Environment* env = self->env;
   // request shutdown is called from reactions which are not executed in critical sections.
   // Thus we enter a critical section before setting the stop tag.
   MUTEX_LOCK(self->mutex);
-  self->stop_tag = lf_delay_tag(self->current_tag, 0);
+  self->stop_tag = shutdown_time;
```
sbgaia: In the non-federated scenario, this logic allows a second call to update an existing stop event if the new shutdown time is greater than the current one (or cancel if the second call uses NEVER as shutdown_time). Is this intentional, or should a shutdown request be strict/immutable once it's requested?

sbgaia (changes requested):
Great work! I left a couple of notes just to double-check some of the logic.

include/reactor-uc/macros_internal.h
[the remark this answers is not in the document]
tanneberger: done

src/shutdown_coordinator.c
tanneberger: true, good catch!

src/shutdown_coordinator.c
tanneberger: The intended behavior is that, the earliest message (in logical time) with the latest logical shutdown time, should win.

tanneberger: I think you are right this would need to be: 

```c
bool announcement_time_equal = (bool)(lf_tag_compare(announcement_tag, self->announcement_of_shutdown) <= 0 ||
                                        lf_tag_compare(self->announcement_of_shutdown, NEVER_TAG) == 0);
                                        
```

for it to actually be smaller or equal otherwise we could never overwrite a shutdown time.

tanneberger:
```c
  bool shutdown_larger = lf_tag_compare(shutdown_tag, self->proposed_shutdown_time) == 1;
  bool announcement_time_less = lf_tag_compare(announcement_tag, self->announcement_of_shutdown) < 0;
  bool announcement_time_equal = lf_tag_compare(announcement_tag, self->announcement_of_shutdown) == 0;

  if (current_step <= self->longest_path && ((announcement_time_equal && shutdown_larger) || announcement_time_less)) {
    LF_INFO(FED, "New and larger shutdown time received! Adjusting the shutdown time and broadcasting it.");
    self->proposed_shutdown_time = shutdown_tag;
```

I think this is now correct: we overwrite the shutdown time IF:
- the announcement is from an earlier logical time
- the announcement is from the same logical time, but has a larger shutdown time

src/schedulers/dynamic/scheduler.c
tanneberger: Good question. For the federated use case, we need the functionality to allow it to be overwritten. Perhaps I can simply add an extra boolean flag to this function (`overwrite`), which is set to true in the federated environment and false in the non-federated environment.

examples/patmos/s4noc_fed/common_config.h
```diff
@@ -20,6 +20,9 @@
 // Startup coordinator: number of startup event slots
 #define STARTUP_EVENT_SLOTS 6
 
+// Shutdown coordinator: number of startup event slots
```
edwardalee, line 23:
```suggestion
// Shutdown coordinator: number of shutdown event slots
```

external/proto/message.pb.h
```diff
@@ -79,6 +79,19 @@ typedef struct _StartupCoordination {
...
+    Tag shutdown_tag;
+    Tag current_tag;
+} ShutdownTimeAnnouncement;
+
+typedef struct _ShutdownCoordination {
+    pb_size_t which_message;
+    union {
+        ShutdownTimeAnnouncement shutdown_time_announcement;
+    } message;
+} ShutdownCoordination;
```
edwardalee, line 93: Is there a reason for the underscore in the struct tag? I think it is more common to make the struct tag the same as the typedef name. I.e., do this:
```
typedef struct ShutdownCoordination {
    pb_size_t which_message;
    union {
        ShutdownTimeAnnouncement shutdown_time_announcement;
    } message;
} ShutdownCoordination;
```
It's not a big deal, but leading underscores can be problematic in C if name collisions occur.

include/reactor-uc/environment.h
```diff
@@ -116,13 +116,17 @@ struct Environment {
...
-   * Any reaction triggered by the shutdown trigger will be executed before the program terminates.
-   * If the program is not federated, then the shutdown will occur at the next microstep.
-   * If the program is federated, then the shutdown tag will be negotiated with the other federates.
+   * This function requests shutdown of the program at the current logical time plus
+   * the specified shutdown offset. Any reaction triggered by the shutdown trigger
+   * will be executed before the program terminates (last tag before shutdown).
+   * If the program is not federated, then a shutdown_offset of 0 causes shutdown
+   * at the next microstep, while a positive shutdown_offset causes shutdown at a
+   * future logical time. If the program is federated, then the resulting shutdown tag will be
+   * announced to all other federates.
```
edwardalee, line 127: This does not make it clear what happens if another federate is already past the proposed shutdown time. What happens?  How is this handled?

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcShutdownCoordinatorGenerator.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class UcShutdownCoordinatorGenerator(
+    private val federate: UcFederate,
+    private val connectionGenerator: UcConnectionGenerator,
+) {
+
+  companion object {
+    // The number of system events allocated for each neighbor. Used to schedule received messages
+    // as system events. The worst-case number of events is 1 per Networkchannel.
+    val numSystemEventsPerBundle = 3
```
edwardalee, line 11: This description doesn't mean anything to me. Are "system events" defined somewhere?

test/lf/src/legacy/Timeout.lf
```diff
@@ -41,9 +41,11 @@ reactor Consumer {
 
 @platform("native")
 @timeout(11ms)
+@logging("Debug")
 main reactor Timeout {
   consumer = new Consumer()
   producer = new Sender(break_interval = 1 msec)
-
+  
+  @maxwait(1000ms)
```
edwardalee, line 49: Why is maxwait an attribute of a connection? Doesn't this regulate advancement of time? If so, it should be an attribute of the destination reactor.

edwardalee (changes requested):
I'm missing any description of how distributed shutdown works. Perhaps this should go in `coordinated_shutdown.h`?

Also, the federated tests all seem to request shutdown 0 ms into the future, which seems unlikely to work.  In at least one case, you added `after` delays, which I guess will help, but  where is the explanation of this?

external/proto/message.pb.h
tanneberger: thats a file generated by protobuf. I think the user should use the typedef name at the end and not the struct name, hence the struct name gets the prefix _.

include/reactor-uc/environment.h
tanneberger: This is an incredibly hard problem. Some thoughts:

Should this federate shut down immediately after receiving the shutdown request? What if there is a second shutdown message for a later time, which would invalidate the first request? Then the program would shut down earlier than needed, and this behavior would depend entirely on the order of messages, making the shutdown non-deterministic.

At startup, this is not an issue because we can ensure that all messages are received before committing to a start time. However, for shutdown, we don't know if more messages are coming in.

Practically speaking, the programmer needs to ensure that the shutdown time is sufficiently far in the future and/or that the `@maxwait` (and potentially after-delay) values are high enough so this message can propagate through the entire federation before the time is reached.

Alternatively, we could implement something more complicated, like a two-wave flooding/gossip protocol that ensures every federate is reached and only on the second pass activates or enables the shutdown. The first wave would set the shutdown value, and the second wave would prime it. However, I would keep it simple for now; if we later realize we need this, we can implement it then.

For this PR I would implement the semantics that if the shutdown time already has passed, that shutdown  happens at the next tag.

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcShutdownCoordinatorGenerator.kt
tanneberger: Currently, although not formalized, the dynamic scheduler has two event queues: one for LF events (such as Actions and Delayed Connections) and a second for events from the runtime itself, like Received Messages and Clock-Sync messages.

The reason for this is that receiving network messages happens asynchronously via callbacks inside the `NetworkChannel`s. Often, this asynchronous receive is triggered by an interrupt routine, and the logic inside those IRQ handlers should be minimal. Therefore, we create a system event and place it on the event queue for the scheduler to handle.

test/lf/src/legacy/Timeout.lf
tanneberger: In reactor-uc the `@maxwait` is a property of the NetworkChannel, hence we annotate in on top-level federated lf connections. (Remember the difference in semantics between reactor-uc and reactor-c maxwait)

include/reactor-uc/environment.h
edwardalee: At a minimum, this should document what happens in this PR. I think the "right" solution is probably a two-wave gossip protocol, but I agree that this is pretty complicated and probably should be left for subsequent work. Perhaps the documentation at this point should say that the handling of this situation may change in the future.

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcShutdownCoordinatorGenerator.kt
edwardalee: Document this somewhere? It seems like a key design decision, and the code will be hard to understand without some clues.

src/schedulers/dynamic/scheduler.c
tanneberger: => if overwrite == true flag, if the shtudown_time != NEVER_TAG => error

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcShutdownCoordinatorGenerator.kt
tanneberger: @edwardalee I added some docs here: https://micro-lf.org/design/scheduling/#system-events

include/reactor-uc/environment.h
tanneberger: I made an issue for it #361

src/shutdown_coordinator.c
```diff
@@ -0,0 +1,145 @@
...
+#include <reactor-uc/logging.h>
+
+#define NUM_RESERVED_EVENTS 3 // 3 events is reserved for scheduling our own events.
+
+/** Handle an incoming message from the network. Invoked from an async context in a critical section. */
+static void ShutdownCoordinator_handle_message_callback(ShutdownCoordinator* self, const ShutdownCoordination* msg,
+                                                        size_t bundle_idx) {
+  LF_DEBUG(FED, "Received shutdown message from neighbor %zu. Scheduling as a system event", bundle_idx);
+  ShutdownEvent* payload = NULL;
+  lf_ret_t ret = self->super.payload_pool.allocate(&self->super.payload_pool, (void**)&payload);
```
sbgaia, line 14: self->super.payload_pool.allocate_reserved?

sbgaia (approved):

src/shutdown_coordinator.c
tanneberger: why do we actually have this functions its never used? 

I would merge this now.

Issue #136, opened 2024-11-28
Try to get code coverage also from the LF tests


erlingrj:
Currently we have integrated lcov with our unit test framework Unity. So we get a coverage report based on all the unit-tests. However, most of the tests are now LF tests and it would be tremendously helpful to also see the coverage of the LF tests

PR #346, opened 2026-04-14
Optimize LF test compilation with CMake

### LINKED ISSUE #275: Reduce test compilation times
After merging #268 it takes around 10minutes to run all the tests. This time could likely be reduced if we did not recompile then runtime for each test but compiled it and nanopb into static archives and linked them with the code-generated code.

sbgaia:
This PR reworks how LF and unit tests are built and executed. Instead of compiling reactor-uc from scratch for each LF test via the Makefile, we now use CMake to share a single reactor-uc build across all tests and compile them in parallel.

I also added a batch processing function in `lfc.cmake` so we can invoke LFC once for multiple test files instead of calling it repeatedly, and added coverage support for LF tests (previously we only had it for unit tests).

Minor changes:
- Moved federated LF tests into `test/lf/src/federated/` with a dedicated runner script;
- Improved runtime symlink handling in `UcPlatformGenerator.kt`;
- Fixed a couple of ASAN errors in connection logging and the event payload pool test;
- Cleaned up the build config so unit tests and LF tests are properly decoupled.

Closes #275 and #136.

### Benchmark

This is the improvement I got on my laptop:

| Approach | Time |
|----------|------|
| Makefile (compiles reactor-uc per test, sequential) | 1210s |
| CMake (shared reactor-uc, `-j16`) | 182s |
| **Speedup** | **6.6x** |

CMake breakdown:
- LFC generation (cmake configure): ~101s
- Compilation: ~18s
- Test execution (ctest): ~63s

We can squeeze more out of this:
1. **Batch generation in LFC for federates**: LFC reports an error if the batch contains more than one main federate. Currently, we still must process each federate individually, adding a lot of overhead.
2. **Move re-generation checks inside LFC**: Currently, when a `.lf` file has already been built, LFC re-executes the entire generation pipeline and checks for each file whether the output it wants to write matches what's already on disk. Instead, we could take the same approach as `make` and check whether the output directory was modified more recently than the `.lf` file and its imports. This would allow us to cache the generated output in the CI pipeline and skip the generation step entirely when the tests haven't changed.


edwardalee (approved):
LGTM, but I did not check it carefully. I would say, if it works, and you are sure all tests are running, go for it.  I would be curious to know how much it speeds up the CI tests.

cmake/lfc.cmake
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.20.0)
 
+set(LFC_RUNTIME_SYMLINK OFF CACHE BOOL "Use a symlink to the runtime instead of copying it into src-gen")
```
tanneberger, line 3: do you need a symlink to the lfc or to reactor-uc?

Is it possible to reuse `REACTOR_UC_PATH` ?

cmake/lfc.cmake
```diff
@@ -44,6 +49,33 @@ function(lf_run_lfc LF_SOURCE_DIR LF_MAIN)
 
 endfunction()
 
+# Run the LFC compiler on multiple LF source files in a single invocation.
+# This avoids paying JVM startup cost per file.
+# Args:
+#   OUTPUT_DIR: The root output directory for generated code.
+#   ARGN: List of absolute paths to .lf files.
+function(lf_run_lfc_batch OUTPUT_DIR)
```
tanneberger: Would it be possible to also make the normal `lf_lfc_run` function variadic, so we can pass an arbitrary amount of files into this function?

I am not a huge fan of duplicating the logic here.

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcPlatformGenerator.kt
```diff
@@ -91,18 +91,28 @@ abstract class UcPlatformGenerator(protected val generator: UcGenerator) {
...
+      runtimeDestinationPath.deleteIfExists()
+    }
+
+    val runtimeArtifacts = listOf("src", "include", "external", "cmake", "make", "CMakeLists.txt")
+
     if (fileConfig.runtimeSymlink) {
-      if (runtimeDestinationPath.exists() && !runtimeDestinationPath.isSymbolicLink()) {
-        runtimeDestinationPath.deleteRecursively()
+      Files.createDirectories(runtimeDestinationPath)
+      for (entry in runtimeArtifacts) {
```
tanneberger, line 104: can you explain this? 

so this basically deletes copied runtime files?

src/connection.c
```diff
@@ -30,7 +30,7 @@ Port* Connection_get_final_upstream(Connection* self) {
 }
 
 void Connection_register_downstream(Connection* self, Port* port) {
-  LF_DEBUG(CONN, "Registering downstream %s.%p with connection %s.%p", port, self);
+  LF_DEBUG(CONN, "Registering downstream %p with connection %p", port, self);
```
tanneberger: I see there are missing arguments. 

I think its better to add the name of the containing reactor back.


something like this:
```c
LF_DEBUG(CONN, "Registering downstream %s.%p with connection %s.%p", port->parent->name, port, self->parent->name, self);
```

test/lf/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+cmake_minimum_required(VERSION 3.20)
+project(LFTests LANGUAGES C)
+
+# Reactor-UC path: either from env or passed via -D
```
tanneberger: REACTOR_UC_PATH

tanneberger: via compile flag -DREACTOR_UC_PATH

test/lf/lf_test_functions.cmake
```diff
@@ -0,0 +1,66 @@
...
+#   - reactor-uc target exists
+#   - LF_COVERAGE_DEPS global property is defined
+#   - CTest is included
+
+set(LF_TEST_DIR ${CMAKE_CURRENT_LIST_DIR})
+set(LF_TEST_TIMEOUT 30 CACHE STRING "Default timeout in seconds for LF tests")
+
+# Build one executable from a generated Include.cmake directory.
+# Optional extra arguments are added as additional include directories.
+function(register_lf_target TARGET_NAME SRC_GEN_DIR)
```
tanneberger: so this function basically adds an executable to cmake and links all the libraries?

Maybe can put this function actually into the generic cmake function file and reuse it here:
https://github.com/lf-lang/reactor-uc/blob/main/lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt#L55

And I think we need a better name: e.g. `lf_add_executable` or something like this.

tanneberger (changes requested):

CMakeLists.txt
```diff
@@ -110,11 +114,39 @@ endif()
...
+if(BUILD_UNIT_TESTS)
   add_subdirectory(test/unit)
 endif()
+
+if(BUILD_LF_TESTS)
+  include(cmake/lfc.cmake)
+  include(test/lf/lf_tests.cmake)
+endif()
+
+if(TEST_COVERAGE)
```
tanneberger: why moving the coverage here? I think it was nicer to not litter our main CMakefile

cmake/lfc.cmake
sbgaia: To use `REACTOR_UC_PATH`, we need to update the LFC compiler. Currently, the generated sources expect a reactor-uc directory inside the generated directory. I suggest we create a separate issue and handle this in next PR.

cmake/lfc.cmake
sbgaia: Done it, now we have a single function. Now the template repo fails. After this PR is merged, I will fix them.

src/connection.c
sbgaia: done

test/lf/CMakeLists.txt
sbgaia: done

CMakeLists.txt
sbgaia: Done, I moved the coverage configuration inside the test folder

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcPlatformGenerator.kt
sbgaia: Previously, the symlink pointed to the entire `reactor-uc` folder, which unintentionally made other directories accessible in the generated output. This change standardizes the behavior between the copy and symlink approaches.

tanneberger (approved):

PR #363, opened 2026-05-20
Add support for Ports of Type void

tanneberger:
Changes to the user API:

Now valid:
```
input x: void
input x
```

```
reaction(t) -> out {=
    lf_set(out);
=}
```

If the port is void, `lf_set` is now allowed to take only the port.

This the last big missing piece for our v0.1. release. :checkered_flag:


include/reactor-uc/macros_api.h
```diff
@@ -106,4 +112,14 @@ lf_ret_t lf_schedule_with_value(Action* action, interval_t offset, const void* v
     lf_ret_t __ret = __a->schedule(__a, (offset), (const void*)array);                                                 \
   })
 
+/**
+ * @brief Sets the value of an output port and triggers all downstream reactions.
+ *
+ * This macro copies the value to the output port and consequently triggers all downstream reactions.
```
edwardalee:
```suggestion
 * This macro copies the value to the output port and consequently triggers all downstream reactions.
 * The value argument need not be given if the port is of type `void`.
```

include/reactor-uc/macros_api.h
```diff
@@ -74,9 +74,15 @@ lf_ret_t lf_schedule_with_value(Action* action, interval_t offset, const void* v
...
 #define GET_ARG4(arg1, arg2, arg3, arg4, ...) arg4
 
+/// @private
+#define GET_ARG3(arg1, arg2, arg3, ...) arg3
+
 /// @private
 #define LF_SCHEDULE_CHOOSER(...) GET_ARG4(__VA_ARGS__, lf_schedule_with_val, lf_schedule_without_val)
 
+/// @private
+#define LF_SET_CHOOSER(...) GET_ARG3(__VA_ARGS__, lf_set_with_val, lf_set_without_val)
```
edwardalee, line 84: I don't understand how this works. Won't this always choose `lf_set_without_val`?

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
```diff
@@ -70,12 +70,9 @@ abstract class UcCmakeGenerator(
...
 open class UcCmakeGeneratorNonFederated(
     private val mainDef: Instantiation,
+    buildType: BuildTypeType.BuildType,
     fileConfig: UcFileConfig,
-) :
-    UcCmakeGenerator(
-        UcLoggingLevelAttribute(mainDef.reactor),
-        BuildTypeType.BuildType.RELEASE,
-        fileConfig) { // FIXME:
+) : UcCmakeGenerator(UcLoggingLevelAttribute(mainDef.reactor), buildType, fileConfig) { // FIXME:
```
edwardalee: What is the FIXME for?

src/schedulers/dynamic/scheduler.c
```diff
@@ -407,6 +407,8 @@ lf_ret_t Scheduler_schedule_at(Scheduler* super, Event* event) {
   DynamicScheduler* self = (DynamicScheduler*)super;
   lf_ret_t ret;
 
+  LF_INFO(SCHED, "Scheduling Event Value: %p", event->super.payload);
```
edwardalee: Should this be `LF_DEBUG`?

edwardalee (approved):
Looks great!  I left some minor comments. It's not obvious to me how the macro magic works, so some hints in the comments would be helpful.

lfc/core/src/main/kotlin/org/lflang/generator/uc/attribute/UcLoggingAttribute.kt
```diff
@@ -18,7 +18,9 @@ class UcLoggingLevelAttribute(var level: String = UcLoggingLevel.INFO) {
   }
 
   fun getLogLevelFromNode(inst: Reactor): String {
-    return when (AttributeUtils.getLoggingAttrValue(inst)) {
+    val annotation = AttributeUtils.getLoggingAttrValue(inst).capitalize()
```
sbgaia: AttributeUtils.getLoggingAttrValue(inst).uppercase()

src/connection.c
```diff
@@ -48,7 +48,10 @@ void LogicalConnection_trigger_downstreams(Connection* self, tag_t intended_tag,
 
     if (down->effects.size > 0 || down->observers.size > 0) {
       validate(value_size == down->value_size);
-      memcpy(down->value_ptr, value, value_size); // NOLINT
+
+      if (value_size > 0) {
```
sbgaia, line 52: Should we add the same check here, too?
src/port.c:32 
src/federated.c:63 
src/federated.c:128 
src/action.c:18 
src/connection.c:163

sbgaia (approved):

src/schedulers/dynamic/scheduler.c
tanneberger: I removed it

include/reactor-uc/macros_api.h
tanneberger: it select either lf_set_with_val or lf_set_without_val based on the number of arguments. 

This "hack" is pretty well explained here: https://stackoverflow.com/questions/679979/how-to-make-a-variadic-macro-variable-number-of-arguments

lfc/core/src/main/kotlin/org/lflang/generator/uc/UcCmakeGenerator.kt
tanneberger: the buildType variable, we need to either have to remove it or fetch it from the annotation.

src/connection.c
tanneberger: done

tanneberger:
document the CHOOSER macros.

PR #379, opened 2026-06-24
.ulf file extension

tanneberger:
This PR:

- changes the accepted file extension from .lf to .ulf
- removes the target uC declaration.


ulf/core/src/main/kotlin/org/lflang/generator/uc/UcFederatedProjectTemplateGenerator.kt
EhsanKhodadad: This line needs to be change to ulf as well.

test/lf/src/legacy/failing/FileReader.ulf
```diff
@@ -1,5 +1,4 @@
 /** Test reading a file at a location relative to the source file. */
-target uC {
```
edwardalee, line 2: This change doesn't look quite right as it leaves:
```
  platform: Native
}
```

test/lf/src/legacy/failing/MethodMacros.ulf
```diff
@@ -1,5 +1,4 @@
 // Test ability of methods to call each other and (recursively) themselves.
-target uC {
```
edwardalee, line 2: Same problem. Why don't these cause test failures?

edwardalee (changes requested):
Looks OK to me, except for a large number of places where the target was not completely removed. Why don't these cause test failures?  I.e., only the first line of the following is being removed:
```
target uC {
  platform: Native
}
```
Also, it would be good to leave at least one test that preserves the `target uC` line to ensure that it remains optional.

Also, there are test failures...

sbgaia (approved):
Looks good to me!

test/lf/src/legacy/failing/MethodMacros.ulf
sbgaia: The tests in the `legacy/failing' folder are currently disabled in our CI pipelines. They look like older `reactor-c' tests that may contain unsupported features. I think we should open a separate issue to investigate and enable the ones that are now supported.

PR #391, opened 2026-07-29
Add zephyr BLE transport channel

sbgaia:
## Overview
This pull request introduces support for Bluetooth Low Energy (BLE) as a new network channel for federated Reactor-UC on Zephyr/Nordic (nRF5x) boards.

## Main Changes
- **Add BLE channel support** for the Zephyr target.
- **Add Kotlin/Java code generation and validation**, along with support for the BLE interface attribute.

## Minor Changes
- **Increase federation joining timeout** to better accommodate BLE latency.
- **Notify the scheduler** when a federate resumes a federation (mimicking the clock-sync behavior).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,6 @@ set(ASAN OFF CACHE BOOL "Compile with AddressSanitizer")
 set(PLATFORM "POSIX" CACHE STRING "Platform to target")
 set(SCHEDULER "DYNAMIC" CACHE STRING "Scheduler to use")
 set(NETWORK_CHANNEL_TCP_POSIX OFF CACHE BOOL "Use POSIX TCP NetworkChannel")
-set(NETWORK_CHANNEL_BLE OFF CACHE BOOL "Use Zephyr BLE NetworkChannel (Nordic nRF5x)")
 set(FEDERATED OFF CACHE BOOL "Compile with federated sources")
 
 
@@ -91,10 +90,6 @@ target_compile_definitions(reactor-uc PRIVATE "SCHEDULER_${SCHEDULER}")
 
 if(NETWORK_CHANNEL_TCP_POSIX)
   target_compile_definitions(reactor-uc PRIVATE NETWORK_CHANNEL_TCP_POSIX)
-endif()
-
-if(NETWORK_CHANNEL_BLE)
-  target_compile_definitions(reactor-uc PRIVATE NETWORK_CHANNEL_BLE)
 endif()
 
 if(FEDERATED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(ASAN OFF CACHE BOOL "Compile with AddressSanitizer")
 set(PLATFORM "POSIX" CACHE STRING "Platform to target")
 set(SCHEDULER "DYNAMIC" CACHE STRING "Scheduler to use")
 set(NETWORK_CHANNEL_TCP_POSIX OFF CACHE BOOL "Use POSIX TCP NetworkChannel")
+set(NETWORK_CHANNEL_BLE OFF CACHE BOOL "Use Zephyr BLE NetworkChannel (Nordic nRF5x)")
 set(FEDERATED OFF CACHE BOOL "Compile with federated sources")
 
 
@@ -90,6 +91,10 @@ target_compile_definitions(reactor-uc PRIVATE "SCHEDULER_${SCHEDULER}")
 
 if(NETWORK_CHANNEL_TCP_POSIX)
   target_compile_definitions(reactor-uc PRIVATE NETWORK_CHANNEL_TCP_POSIX)
+endif()
+
+if(NETWORK_CHANNEL_BLE)
+  target_compile_definitions(reactor-uc PRIVATE NETWORK_CHANNEL_BLE)
 endif()
 
 if(FEDERATED)
--- a/include/reactor-uc/platform/zephyr/ble_channel.h
+++ b/include/reactor-uc/platform/zephyr/ble_channel.h
@@ -108,6 +108,10 @@ struct BleChannel {
   FederateMessage output;
   unsigned char receive_buffer[BLE_CHANNEL_BUFFERSIZE];
   unsigned int receive_buffer_index;
+  // Incremented every time the buffer is flushed on link loss. poll() snapshots it before
+  // scanning and re-checks it under `rx_lock` before compacting, so a flush that lands
+  // mid-poll cannot make it consume offsets that no longer exist.
+  uint32_t rx_generation;
 
   // Send scratch buffer (framed message before chunking onto PDUs).
   unsigned char send_buffer[BLE_CHANNEL_BUFFERSIZE];
--- a/include/reactor-uc/platform/zephyr/ble_channel.h
+++ b/include/reactor-uc/platform/zephyr/ble_channel.h
@@ -1,0 +1,129 @@
+#ifndef REACTOR_UC_ZEPHYR_BLE_CHANNEL_H
+#define REACTOR_UC_ZEPHYR_BLE_CHANNEL_H
+
+/**
+ * @brief A Bluetooth Low Energy NetworkChannel for federated Reactor-UC on
+ * Zephyr / Nordic (nRF5x) boards.
+ *
+ * One channel == one BLE connection between exactly two boards:
+ *   - one board is the PERIPHERAL: it advertises under `device_name` and hosts a
+ *     small GATT service (NUS-compatible UUIDs), then waits for a central;
+ *   - the other board is the CENTRAL: it scans for `device_name`, connects,
+ *     discovers the service, subscribes to notifications and writes to the RX
+ *     characteristic.
+ *
+ * MULTIPLE LINKS PER BOARD: several channels may coexist. A CENTRAL board (e.g.
+ * a fusion federate) can hold one channel per upstream PERIPHERAL, each with its
+ * own `device_name` and connection parameters (not tested yet). 
+ * The runtime keeps a registry and routes every BLE event to the owning 
+ * channel. (A single board should host at most one PERIPHERAL channel, 
+ * because legacy advertising exposes one device name at a time. Set 
+ * CONFIG_BT_MAX_CONN to the number of links the board needs.)
+ *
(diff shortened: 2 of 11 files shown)


include/reactor-uc/platform/zephyr/ble_channel.h
```diff
@@ -0,0 +1,129 @@
...
+ *     discovers the service, subscribes to notifications and writes to the RX
+ *     characteristic.
+ *
+ * MULTIPLE LINKS PER BOARD: several channels may coexist. A CENTRAL board (e.g.
+ * a fusion federate) can hold one channel per upstream PERIPHERAL, each with its
+ * own `device_name` and connection parameters (not tested yet).
+ * The runtime keeps a registry and routes every BLE event to the owning
+ * channel. (A single board should host at most one PERIPHERAL channel,
+ * because legacy advertising exposes one device name at a time. Set
+ * CONFIG_BT_MAX_CONN to the number of links the board needs.)
```
edwardalee, line 21: This is not clear to me and appears contradictory. What is `CONFIG_BT_MAX_CONN`? An environment variable? There is no further mention of it in this file.  Do I infer that you can have multiple links per channel?

edwardalee (approved):
This looks great to me, but I'm no expert on BLE. I've requested a Copilot review.

My major concern is the lack of testing. This would be a great candidate for @hokeun 's ACT testing framework. Could we make a regression test that runs on a pair of boards, supervised by a self-hosted runner?

include/reactor-uc/platform/zephyr/ble_channel.h
sbgaia: `CONFIG_BT_MAX_CONN` is a Zephyr configuration variable that defines the maximum number of simultaneous Bluetooth connections the stack supports. It is typically set via environment configuration (`prj.conf` / `-DCONFIG_...`).

Yes, you can have multiple links per peripheral channel. A single board should only host one peripheral channel (because legacy advertising exposes one device name at a time), but a central device can connect to multiple peripherals simultaneously. `CONFIG_BT_MAX_CONN` just needs to match the total number of simultaneous links the board will maintain.

Currently, I have only tested this implementation with 2 boards (a simple 1-to-1 setup). In theory, this implementation supports multiple boards and links, but I am still waiting for my third board to arrive. Temporarily, I am forcing `CONFIG_BT_MAX_CONN` to 1 [here](https://github.com/lf-lang/reactor-uc/pull/391/changes#diff-34ab40f7e64bbb278595da2eb6da8cdf8f885924e64a8dbf0f66c75407527144R51).

include/reactor-uc/platform/zephyr/ble_channel.h
```diff
@@ -0,0 +1,129 @@
...
+  // Receive buffering (filled in BLE callback context, drained in poll()).
+  FederateMessage output;
+  unsigned char receive_buffer[BLE_CHANNEL_BUFFERSIZE];
+  unsigned int receive_buffer_index;
+
+  // Send scratch buffer (framed message before chunking onto PDUs).
+  unsigned char send_buffer[BLE_CHANNEL_BUFFERSIZE];
+
+  FederatedConnectionBundle* bundle;
+  void (*receive_callback)(FederatedConnectionBundle* bundle, const FederateMessage* message);
```
tanneberger, line 120: I I think you can drop this and instead use the "inherited" one from normal AsyncNetworkChannel.

include/reactor-uc/platform/zephyr/ble_channel.h
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @brief Construct a BLE channel for one federated link.
+ *
+ * @param self        Storage for the channel (lives inside the connection bundle)
+ * @param role        BLE_CHANNEL_ROLE_PERIPHERAL or BLE_CHANNEL_ROLE_CENTRAL
+ * @param device_name Advertised name of the PERIPHERAL end of THIS link (both ends agree)
+ * @param params      Connection interval, slave latency, and supervision timeout for this link
+ */
+void BleChannel_ctor(BleChannel* self, BleChannelRole role, const char* device_name, BleConnParams params);
```
tanneberger: I think its not so cool that this struct leaks - the problem is that we need to let the code generator construct this object, which might be a bit more annoying that having a lot of arguments.

src/startup_coordinator.c
```diff
@@ -10,7 +10,11 @@
 #define NUM_RESERVED_EVENTS 3 // 3 events is reserved for scheduling our own events.
 
 #ifndef TRANSIENT_WAIT_TIME
-#define TRANSIENT_WAIT_TIME MSEC(250)
+// How long a (re)joining federate waits for neighbors to answer its start-time
+// request before giving up. 250ms is too tight over a real link (e.g. a 30ms BLE
+// connection just re-established, still draining buffered traffic) -> the join
+// times out and the federate shuts down. 2s leaves ample margin.
+#define TRANSIENT_WAIT_TIME MSEC(2000)
```
tanneberger: 2s is not pretty long for other Network Channels we should consider making this parameterizable.

src/startup_coordinator.c
```diff
@@ -449,20 +459,29 @@ static void StartupCoordinator_handle_start_time_response(StartupCoordinator* se
   self->start_time_proposal = start_time;
   self->state = StartupCoordinationState_RUNNING;
 
-  instant_t joining_time = 0;
+  instant_t joining_time = max_logical_time + MSEC(50);
```
tanneberger, line 458: yeah what we dicussed in the call: consider cherry picking that clock-sync happens before starup.

ulf/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
```diff
@@ -224,6 +235,34 @@ class UcS4NocInterface(val core: Int, name: String? = null) :
...
+
+  fun createEndpoint(): UcBleEndpoint {
+    val ep = UcBleEndpoint(deviceName, interval, latency, timeout, this)
+    endpoints.add(ep)
+    return ep
+  }
+
+  companion object {
+    fun fromAttribute(federate: UcFederate, attr: Attribute): UcBleInterface {
+      val deviceName = attr.getParamString("device_name") ?: "reactor-uc-ble"
```
tanneberger: that is  a valid device-name? I though thats something inside zephyr?

ulf/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
```diff
@@ -224,6 +235,34 @@ class UcS4NocInterface(val core: Int, name: String? = null) :
...
+      val interval = attr.getParamInt("interval") ?: 30
```
tanneberger: maybe I am fully hallucinating: but isn't there a time value/type that you could use?

ulf/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
```diff
@@ -411,6 +455,24 @@ class UcS4NocChannel(
...
 
+class UcBleChannel(
+    private val ble_src: UcBleEndpoint,
+    private val ble_dest: UcBleEndpoint,
+    serverLhs: Boolean = true,
+) : UcNetworkChannel(BLE, ble_src, ble_dest, serverLhs) {
+
+  // The source (LHS of `->`) is the BLE peripheral when serverLhs is true.
+  private fun ctor(ep: UcBleEndpoint, peripheral: Boolean) =
+      "BleChannel_ctor(&self->channel, ${if (peripheral) "BLE_CHANNEL_ROLE_PERIPHERAL" else "BLE_CHANNEL_ROLE_CENTRAL"}, \"${ep.device_name}\", ((BleConnParams){.interval = BLE_CI_UNITS(${ep.interval}), .latency = ${ep.latency}, .supervision_timeout = BLE_TIMEOUT_UNITS(${ep.timeout})}));"
```
tanneberger: yeah so I see so you create the object inline.

CMakeLists.txt
```diff
@@ -92,6 +93,10 @@ if(NETWORK_CHANNEL_TCP_POSIX)
   target_compile_definitions(reactor-uc PRIVATE NETWORK_CHANNEL_TCP_POSIX)
 endif()
 
+if(NETWORK_CHANNEL_BLE)
```
tanneberger: I am not sure if this is necessary, because the reason the TCP/IP channel is here, so we want the easy compile method directly from the compiler, but for BLE you have to go through the Zephyr cmake anyway.

tanneberger (changes requested):

src/startup_coordinator.c
sbgaia: reverted back to 250, I think would be better to tackle this in a separate PR in which we may analyze the behavior of heterogeneus channels

include/reactor-uc/platform/zephyr/ble_channel.h
sbgaia: Fixed, now we set all the parameters

src/startup_coordinator.c
sbgaia: I have opened another PR, where this is now fixed: #396

ulf/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
sbgaia: Yes, but it can be changed with `bt_set_name()` by setting `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` in the zephyr prj.conf. Alternatively we would have to set it with ` CONFIG_BT_DEVICE_NAME` in the zepyr prj.conf. Personally I prefer having it inside LF, since we are setting the BLE connection parameters here.

ulf/core/src/main/kotlin/org/lflang/generator/uc/UcNetworkChannel.kt
sbgaia: You are right, now this is fixed

CMakeLists.txt
sbgaia: You are right, I reverted this change.

tanneberger (approved):

Issue #392, opened 2026-08-03
Multiports violate assertions in federated environments and cannot access value of multiport via `self->multiport[i]->value`


NicoBiernat:
# Multiports violate assertions in federated environments

#### micro-LF/reactor-uc:
- current main-branch (04bf33ffda94acc7ae88795e1d6ad6f423ec6379)  
#### Tested platforms:
- POSIX (Arch Linux)
- Zephyr OS (v4.4.0) on W5500-EVB-Pico (with custom bootloader, config and init-code that should not affect the LF runtime)

#### Toolchain
- CMake 4.4.1
- GCC 16.1.1
- Zephyr SDK 1.0.1

## Example LF program using Multiports:
```
reactor Sender(n: size_t = 2) {
    output[n] out: int
    timer t(0, 1s)

    reaction(t) -> out {=
        for (size_t i = 0; i < self->n; i++) {
            lf_set(out[i], i);
            printf("Sent %d to multiport index %d\n", i, i);
        }
    =}
}

reactor Receiver(n: size_t = 2) {
    input[n] in: int

    reaction(in) {=
        for (size_t i = 0; i < self->n; i++) {
            if (lf_is_present(in[i])) {
                int val = in[i]->value;
                printf("Received %d on multiport index %d\n", val, i);
            }
        }
    =}
}

federated reactor {
    recv = new Receiver(n = 2)
    send = new Sender(n = 2)

    send.out -> recv.in
}
```

Compilation succeeds, but the following assertions are violated at runtime:  
#### In function Connection_register_downstream:
https://github.com/lf-lang/reactor-uc/blob/04bf33ffda94acc7ae88795e1d6ad6f423ec6379/src/connection.c#L35

#### In function Reactor_validate:
https://github.com/lf-lang/reactor-uc/blob/04bf33ffda94acc7ae88795e1d6ad6f423ec6379/src/reactor.c#L38

#### Federate "Receiver" on POSIX (Linux):
```
$ build/MultiportFederatedBug/recv/MultiportFederatedBug_recv
Assertion failed at  /home/nico/git/master-thesis/lf-posix-uc/src-gen/MultiportFederatedBug/recv/reactor-uc/src/connection.c:35
```

#### Federate "Sender" on POSIX (Linux):
```
$ build/MultiportFederatedBug/send/MultiportFederatedBug_send
[...]

PR #394, opened 2026-08-06
ulfc: fixing bug, that grouping happend not based on source port

tanneberger:


NicoBiernat (approved):
This fixes the first part of #392, good catch! 👍

PR #396, opened 2026-08-14
Fix federation and clock-synchronization bugs

sbgaia:
A federate learns which neighbor is its clock-sync master from a priority message. The grandmaster announced its priority exactly once, from `ClockSynchronization_ctor`, and the only other trigger for a broadcast is a node's own priority changing, which never happens for the grandmaster. Nothing hooked channel reconnection either.

So any non-grandmaster that started, or restarted, after that one announcement came up with `neighbor_clock[i].priority = UNKNOWN_PRIORITY` and `master_neighbor_index = -1`, and its periodic sync round hit the dead branch every period forever.

To solve this we now send a `ClockSyncMessage_priority_request_tag`. The no-master branch now pulls the priorities instead of idling, which activates that path. The periodic round already reschedules itself, so retry is free: the node asks once per period until a neighbor answers.

Other fixes in this branch

- `EventQueue_remove` keeps the heap invariant when the relocated element is smaller than its new parent; event removal and replacement are mutex-protected.
- `lf_time_add / lf_tag_add` saturate instead of overflowing, time macros parenthesize their arguments.
- `PhysicalClock` gained `step_time` for relative adjustments, and clamps elapsed intervals so a backwards platform-clock step cannot flip the sign of the frequency adjustment.
- Clock offsets above the step threshold are stepped whenever they occur, not only on the first round, so a link returning after an outage recovers in one round instead of slewing for minutes.
- The master is no longer demoted on a transient send_blocking failure, which previously orphaned a node on any reconnect window.
- Event-queue tag shifting moved from the scheduler into the queue, under a single lock, with a reheapify afterwards. Clamping at zero is not injective, so a large backwards shift can invert the relative order of two events.
- Payload is freed when a delayed connection fails to schedule.
- Unit tests for tag/time operations, clock stepping, and shift_all_tags.


include/reactor-uc/physical_clock.h
```diff
@@ -30,6 +30,18 @@ struct PhysicalClock {
...
+   * @brief Step the clock by a relative amount.
+   *
+   * Unlike a `get_time` followed by `set_time`, the read and the write happen under a
+   * single acquisition of the clock mutex, so the underlying hardware clock cannot
+   * advance in between and the clock is shifted by exactly @p step.
+   *
+   * @return LF_INVALID_VALUE if the step would drive the clock negative, in which case
+   *         the clock is left untouched. LF_OK otherwise.
+   */
+  lf_ret_t (*step_time)(PhysicalClock* self, interval_t step);
```
tanneberger, line 43: So this is basically for "atomically" updating the offset?

include/reactor-uc/queues.h
```diff
@@ -35,12 +35,49 @@ struct EventQueue {
...
   /** @brief Restore the heap invariant downward from @p idx. */
   void (*heapify)(EventQueue* self, size_t idx);
-  /** @brief Find an event with the same tag and trigger as @p event, or NULL if not found. */
+  /**
+   * @brief Add @p step to the tag of every queued event, clamping at zero.
+   *
+   * Used when the physical clock is stepped, so that pending events keep their intended
+   * position relative to the corrected clock.
+   */
+  void (*shift_all_tags)(EventQueue* self, interval_t step);
```
tanneberger, line 44: Is this done for all events? or just system events? Because I would be opposed to changing the logical time of normal events.

src/clock_synchronization.c
```diff
@@ -14,32 +14,58 @@ static void ClockSynchronization_correct_clock(ClockSynchronization* self, Clock
   LF_DEBUG(CLOCK_SYNC, "Correcting clock. T1=" PRINTF_TIME " T2=" PRINTF_TIME " T3=" PRINTF_TIME " T4=" PRINTF_TIME,
            timestamps->t1, timestamps->t2, timestamps->t3, timestamps->t4);
   interval_t rtt = (timestamps->t4 - timestamps->t1) - (timestamps->t3 - timestamps->t2);
+  // A negative round-trip time is physically impossible. It indicates message reordering,
+  // a clock step on the peer in the middle of the round, or a corrupted timestamp. Discard
+  // the round instead of feeding a bogus sample into the servo.
```
tanneberger, line 19: servo?

include/reactor-uc/queues.h
sbgaia: Yes, this is done only for systems events.

src/environments/federated_environment.c
```diff
@@ -61,6 +61,21 @@ static interval_t FederatedEnvironment_get_physical_time(Environment* super) {
...
 }
 
+/**
+ * @brief Elapsed physical time, measured on the synchronized timescale.
+ *
+ * The base class implementation reads the raw hardware clock, but `start_time` is on the
+ * synchronized timescale in a federation. Subtracting one from the other would leave an
+ * error equal to the whole accumulated clock-sync correction.
+ */
+static interval_t FederatedEnvironment_get_elapsed_physical_time(Environment* super) {
```
tanneberger, line 71: why add this function I dont see it used anywhere inside this PR - or is this intended for user-land?

src/clock_synchronization.c
```diff
@@ -14,32 +14,58 @@ static void ClockSynchronization_correct_clock(ClockSynchronization* self, Clock
...
-    if (clock_offset_abs > CLOCK_SYNC_INITAL_STEP_THRESHOLD) {
-      LF_INFO(CLOCK_SYNC, "Initial clock offset to grand master is " PRINTF_TIME
-                          " which is greater than the initial step threshold. Stepping clock");
-
-      env_fed->clock.set_time(&env_fed->clock, env_fed->clock.get_time(&env_fed->clock) + clock_offset);
-      // Also inform the scheduler that we have stepped the clock so it can adjust timestamps
-      // of pending events.
-      self->env->scheduler->step_clock(self->env->scheduler, clock_offset);
-      self->env->platform->notify(self->env->platform);
+  // Step the clock (forwards or backwards) whenever the offset is too large to slew away
```
tanneberger, line 29: s/slew/slow

src/clock_synchronization.c
```diff
@@ -14,32 +14,58 @@ static void ClockSynchronization_correct_clock(ClockSynchronization* self, Clock
...
+  // in reasonable time. This is deliberately not restricted to the first round: a master
+  // that steps its own clock, or a link that returns after an outage, produces a large
+  // offset later on too, and slewing that at the servo's clamped rate takes minutes.
+  interval_t clock_offset_abs = clock_offset > 0 ? clock_offset : -clock_offset;
+  if (clock_offset_abs > CLOCK_SYNC_INITAL_STEP_THRESHOLD) {
```
tanneberger, line 34: This I dont understand: Why discard an update the the offset is to large? why would be do this?

src/clock_synchronization.c
```diff
@@ -14,32 +14,58 @@ static void ClockSynchronization_correct_clock(ClockSynchronization* self, Clock
...
+  self->has_initial_sync = true;
 
   // Record last error. Currently unused, but can be used to implement the derivative part of a PID
   self->servo.last_error = clock_offset;
   // Integrate the error, used for the integral part of the PID.
   self->servo.accumulated_error += clock_offset;
+  // Anti-windup: bound the integrator itself, not just the controller output. Without
+  // this the accumulated error keeps growing while the correction is saturated, so the
+  // servo stays saturated long after the error is gone (and can eventually overflow).
+  if (self->servo.accumulated_error > self->servo.clamp) {
```
tanneberger, line 64: servo?

src/physical_clock.c
```diff
@@ -12,44 +12,88 @@ lf_ret_t PhysicalClock_set_time(PhysicalClock* self, instant_t time) {
...
-  LF_DEBUG(CLOCK_SYNC, "Setting physical clock to " PRINTF_TIME " offset is " PRINTF_TIME, time, self->offset);
+  LF_DEBUG(CLOCK_SYNC, "Setting physical clock to " PRINTF_TIME " offset is " PRINTF_TIME, time, offset_snapshot);
   return LF_OK;
 }
 
+/**
+ * @brief Compute the synchronized time corresponding to @p current_hw_time.
+ * The caller must already hold `self->mutex`.
+ */
+static instant_t time_at_locked(PhysicalClock* self, instant_t current_hw_time) {
```
tanneberger, line 27: Maybe prefix it with `PhysicalClock_time_at_locked` so it fits with the OOP inspired naming convention

tanneberger (changes requested):

src/clock_synchronization.c
sbgaia: I update the comment, referencing the `self->servo` used by the PI controller.

src/environments/federated_environment.c
sbgaia: It's an override of the federated environment, which is used by `lf_time_physical_elapsed()`. Previously this was subtracting the raw hardware clock which contained an error equal to the whole clock-sync correction we apply at the start.

src/clock_synchronization.c
sbgaia: I dont know if there is a better word for slew in this context. Slew seems more used in this context (PTP/NTP).

src/clock_synchronization.c
sbgaia: What actually happens here is that we fully apply the offset when it is too large, but we skip the update computed by the PI, since this would imply updating the clock two times.

src/physical_clock.c
sbgaia: done

src/clock_synchronization.c
sbgaia: This was already there, I think it takes inspiration from the linuxptp implementation.

src/clock_synchronization.c
tanneberger: @tanneberger check if servo comes from Linux PTP

PR #399, opened 2026-08-21
Debug the hang problem in Patmos's CI/CD test.

EhsanKhodadad:
This PR aims to solve the problem with Patmos, where the a test on both real hardware and emulator does not terminate.


EhsanKhodadad:
almost all tests passed except macos

PR #402, opened 2026-08-28
Unify UART Channels

sbgaia:
Adds a UART NetworkChannel so federates can talk over a plain two-wire serial link, and restructures the existing pico/RIOT implementations around a shared core.

Major chages:
- **Framing:** New `network_channel/frame.{h,c}`: COBS encoding with 0x00 as the frame delimiter, plus a CRC-32/ISO-HDLC trailer. COBS guarantees the payload contains no zero byte, so a receiver that loses sync recovers at the next delimiter, and the CRC rejects the truncated fragment. LfFrameReceiver consumes bytes one at a time and reports LF_FRAME_OK / ERR_CRC / ERR_DECODE / ERR_OVERFLOW.
- **Shared core:** New `network_channel/uart_channel.{h,c}` holds everything platform-independent: the RX ring the ISR feeds, frame decoding in `poll()`, and the TX transfer helpers. A platform binding now only supplies `write()`, `teardown()` and an ISR that pushes bytes into the ring. That cuts `platform/pico/uart_channel.c` and `platform/riot/uart_channel.c` down substantially and makes the new Zephyr binding small. Zephyr resolves uart_device = N to the devicetree alias lfuartN, so pins stay a board concern. Pico and RIOT keep their fixed mappings (uart0 = GP0/GP1, uart1 = GP8/GP9). TX is interrupt-driven on all three: write() arms the transfer and blocks on a semaphore while the TX interrupt drains it into the hardware FIFO, with a baud-derived timeout.
- **Codegen:** `@interface_uart` gains data_bits, parity, stop_bits, async. `UARTDeviceManager` scopes device reservations per federate, since uart_device = 0 on two federates is two different boards. `@platform` parsing is centralized in `PlatformType.Platform`, which now carries its accepted spellings so parsing and validation can't drift apart.


include/reactor-uc/network_channel/frame.h
```diff
@@ -0,0 +1,93 @@
...
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * @brief CRC-32/ISO-HDLC (IEEE 802.3, same as zlib lf_crc32).
+ *
+ * Reflected, polynomial 0xEDB88320, init 0xFFFFFFFF, final XOR 0xFFFFFFFF.
+ * CRC32("123456789") == 0xCBF43926.
+ */
+uint32_t lf_crc32(const uint8_t* data, size_t len);
```
tanneberger: maybe rename this to uc_crc32 or something.

I think we should step back, we already have a couple of function like the methods that operate on tags, we should harmonize it. Because we are here in the runtime, I think a runtime specific prefix makes the most sense.

